PR #12, opened 2015-06-24
Fixed MSET & FBI installations

dukesrg:
- fixed MSET installation for US
- fixed FBI installation for non-EU
- added preliminary support for CN/KR/TW

P.S. Though I'm still not able to build & check, so check twice before merging with master


roxas75:
I'll wait for merging this pull, since i was already working on it.
Though good work, but i actually need to control the hashes too for mset, becouse installing wrong things is not helpful for sysnand, and i do not want bricks.

dukesrg:
Since support_regions is set to 3, CHN/KOR/TWN code is not reachable. It is just for future use, offcourse MSET hashes for the new regions must be added.

dukesrg:
Moved to new pool request

PR #13, opened 2015-06-24
Shrinked region detection code for MSET & FBI installation

dukesrg:
Additional regions data is exist, but commented.
Need only to add hashes for MSET and uncomment title IDs, the rest of the code is ready.

### DIFF
--- a/rxtools/source/features/downgradeapp.c
+++ b/rxtools/source/features/downgradeapp.c
@@ -123,7 +123,7 @@ void downgradeMSET(){
                 FileWrite(&dg, buf, dgsize, 0);
                 FileClose(&dg);
                 //[/Decrypting]
-    
+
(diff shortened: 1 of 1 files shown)


roxas75:
Can you please rewrite this updates basing them on my new fixed version of downgradeapp.c?
I have difficulties merging.
Also, correct the Healt&Safety app regions, you use 0x223 for J, U and E, while it should be 0x203, 0x213, 0x223

dukesrg:
OK, I'll rebase and apply changes. Well, actually need to fork again because of the lack of git client at work :( Will be done in several minutes anyway.

Syphurith:
@dukesrg Git itself have rebase command, you can figure it if don't want to manually do commits again.

PR #18, opened 2015-06-24
Use python in release tools to support multiplatform

swarzesherz:


roxas75:
Does it work with python 2.7?

swarzesherz:
Yes, tested with Python 2.7.5

roxas75:
Thanks!

dukesrg:
I'd better try to make the project homogenous, i.e. get rid of the python code at all and swith to devkit assembler if possible. Anyway we have a bunch of tools to compile to native code of the build system that requires gcc.

mateusmedeiros:
@dukesrg But isn't forcing the enduser to install Python better than forcing the enduser to compile the tools manually or forcing the enduser to use windows? Sadly it's not always that wine works out-of-the-box.

dukesrg:
Wine is not needed to build. It took about 2 hours for me to make a working toolchain on Windows from the scratch. It will took 10-15 minutes for anyone because all the needed tools with links is now listed on Readme.md.
I think the best way is to provide endusers with a single full-featured companion app as a binary for each platform.

mateusmedeiros:
Yes, but it's like you say, these tools are for the enduser. What the enduser wants is to just download and click, he doesn't want to build stuff. Yes, if he needs to install Python, he won't just download and click, but I personally think it would be less trouble. And, who knows, he might have python installed already, too. 

A single full-featured companion app as a binary for each platform would work as well, as long as the same codebase is kept. Actually, it's not like I really care for Python, what I care is for portability. So, as long as the solution is something cross-platform, that would be great.

Syphurith:
Or just leave the link and let user download those himself?

swarzesherz:
Yes the scrips or binaries only download an url
firmware.bin: http://nus.cdn.c.shop.nintendowifi.net/ccs/download/0004013800000002/00000049
msetdg.bin
JPN: http://nus.cdn.c.shop.nintendowifi.net/ccs/download/0004001000020000/0000001A
USA: http://nus.cdn.c.shop.nintendowifi.net/ccs/download/0004001000021000/0000001F
EUR: http://nus.cdn.c.shop.nintendowifi.net/ccs/download/0004001000022000/00000019

To cross-platform binary this code's should work
```
//msetdg.c
#include <stdio.h>
#include <stdlib.h>
#include <curl/curl.h>

size_t write_data(void *ptr, size_t size, size_t nmemb, FILE *stream) {
    size_t written = fwrite(ptr, size, nmemb, stream);
    return written;
}

int main(void) {
    CURL *curl;
    FILE *fp;
    CURLcode res;
    char FileName[] = "msetdg.bin";
    char urlname[256];
    unsigned int tid[] = { 0x00020000, 0x00021000, 0x00022000}; //JPN, USA, EUR
    unsigned int tver[] = { 0x1A, 0x1F, 0x19};
    int choice = 0;
    do{
#ifdef WINDOWS
        system("cls");
#else
        system("clear");
#endif
        printf("Select your 3DS region :\n    [1] JPN\n    [2] USA\n    [3] EUR\n\nSelection : ");
        scanf("%d", &choice);
    }while(!choice || choice > 3);
    sprintf(urlname, "http://nus.cdn.c.shop.nintendowifi.net/ccs/download/00040010%08X/%08X", tid[choice-1], tver[choice-1]);
    printf("%s\n", urlname);
    printf("Downloading MSET 4.x from the CDN...\n");
    curl = curl_easy_init();
    if (curl) {
        fp = fopen(FileName,"wb");
        curl_easy_setopt(curl, CURLOPT_URL, urlname);
        curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_data);
        curl_easy_setopt(curl, CURLOPT_WRITEDATA, fp);
        res = curl_easy_perform(curl);
        curl_easy_cleanup(curl);
        fclose(fp);
        printf("Downloaded!\n");
    }
    return 0;
}
```
```
//cdn_firm.c
#include <stdio.h>
#include <curl/curl.h>

size_t write_data(void *ptr, size_t size, size_t nmemb, FILE *stream) {
    size_t written = fwrite(ptr, size, nmemb, stream);
    return written;
}

int main(void) {
    CURL *curl;
    FILE *fp;
    CURLcode res;
    char FileName[] = "firmware.bin";
    char urlname[256];
    int id = 0x49;
    sprintf(urlname, "http://nus.cdn.c.shop.nintendowifi.net/ccs/download/0004013800000002/%08X", id);
    printf("%s\n", urlname);
    printf("Downloading 3DS firmware from the CDN...\n");
    curl = curl_easy_init();
    if (curl) {
        fp = fopen(FileName,"wb");
        curl_easy_setopt(curl, CURLOPT_URL, urlname);
        curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_data);
        curl_easy_setopt(curl, CURLOPT_WRITEDATA, fp);
        res = curl_easy_perform(curl);
        curl_easy_cleanup(curl);
        fclose(fp);
        printf("Downloaded!\n");
    }
    return 0;
}
```
But as commented by @mateusmedeiros this tools is for end user and to me is more portable a script in python to one binary by each OS

Syphurith:
I think user can use his downloader, such as aria2, orbit, or other ones, since he knows the links then.

swarzesherz:
Yes @Syphurith or simply download file from url and done (just copy and paste in browser address bar and rename)

Syphurith:
@swarzesherz So that's fine for that purpose. I think i understand it. Maybe some links mentioned in another file, that for those smarters. Or other just stick with their python. All okey then.

PR #21, opened 2015-06-25
Add extra validation to NAND CTR start and refactor

swarzesherz:
TODO
Eval NULL after FindNandCtr

### DIFF
--- a/rxtools/source/lib/fs.c
+++ b/rxtools/source/lib/fs.c
@@ -15,31 +15,34 @@ int GetSystemVersion(){
 	return sysversion;
 }
 
-void InitializeNandCrypto(){
-	u32 ctrStart=0;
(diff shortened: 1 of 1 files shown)


rxtools/source/lib/fs.c
```diff
@@ -15,31 +15,34 @@ int GetSystemVersion(){
...
+{
+    static const u8* version_ctrs[] = {
+        (u8*)0x080D7CAC,//4.x
+        (u8*)0x080D858C,//5.x
+        (u8*)0x080D748C,//6.x
+        (u8*)0x080D740C,//7.x
+        (u8*)0x080D74CC,//8.x
+        (u8*)0x080D794C //9.x
+    };
+    static const u32 version_ctrs_len = sizeof(version_ctrs) / sizeof(u32);
```
Syphurith: Only a note: these static variables act as global ones that occupy a constant position in the binary files. Only the scope is limited to the same file.

PR #24, opened 2015-06-25
MSET ported from armips to gas

dukesrg:
*arm11hax binary differ by several commands because of the *.create* directive workaround, thus rop.bin and mset.bin also differ
*rxinstaller.nds is the same

P.S. before merging need to check if this mset works the same way as original. I have no hardmod and never used mset.


Syphurith:
Thanks for your contribution. Eh, maybe you would change commit names or squash some into one..

dukesrg:
Job is done mostly from the place where only web client is available and clipboard is limited, so ought to push frequently. This pull request is only to indicate that mset ported and should be checked by someone with a hardmod before actual merge. I'll port the rest and then refork from the master and merge changes into one pool request. IIRC only several parts of rxmode code left on armgis, brahma and the rest is already on gas.

roxas75:
If you can test it and it will work i'll merge!

dukesrg:
Can you give me a brief describe how to quick test it?
As I can see, there are only two result binaries from msethax - .nds installed, which is the same, and mset.bin, which is a part of rxTools dat and have only five opcodes different in arm11hax

BTW, porting rxmode patches now, how many are they? It comes ut that I only have 9 in armips toolchain build and 11 in gas toolchain!

roxas75:
Should in theory be 10 in armips.
Also, mset.bin is inserted in rxTools.dat at offset 0x0. From there the rop chain will be loaded from the first-stage rop installed by rxinstaller.

dukesrg:
Yes, I know about the mset offset in rxTools.dat. I assumed the size of mset.bin is fixed, is that right?

Do you mean 10 patches total or not counting the arm9 (0801a4c0)?
Got 11, reboothax was not compiled somehow.
BTW, is reboot section filename and ram address actually differ
*.create build/08094454.bin, 0x080FF600*

roxas75:
10 only from armips, 11 counting 0801a4c0.
Yes, that's a cool trick, since that code gets copied at firm boot from 0x08094454 to 0x080FF600, so i have to compile the code to be executed at 0x080FF600, but i have to insert the patch at 0x08094454.

dukesrg:
The funny thing is that this patch have no ralative addresses so ramaddress does not matter at all. GAS compiles byte equal code for that patch. Only have 4 other patches to port, the rest is OK with GAS!

So about mset, can I check arm11hax is working using 9.2 console? Running rxinstall is not a problem, i have dstwo.

roxas75:
You'll probably have to change the value of **fcram_code_addr** from 0x03E6D000 to 0x03F00000 to make it work on 9.2.

dukesrg:
I'm actually need to check only arm11hax work, not the complete mset

Syphurith:
Maybe you could see the compare results via static analysis, since it is only a ported target.

roxas75:
I like all the work you did, but personally i found very useful to have the name of the patch in the filename, and this is not possible in this conversion in gas.
Just sayin', is it really necessary to delete armips dependency?

dukesrg:
Well, as long as armips don't have an official release published on the source page it will be useful for quick-start. Luckily I have VS2013 installed so it did not took much time to solve one additional dependency and build it. So it's just a matter of a jump-in time for new devs.
We can keep name of the patches by altering the makefile, we just can't keep several patches in the same file with gas.

PR #25, opened 2015-06-26
Fixed IndentationError, NameError

exhalatio:


Syphurith:
Well there are similar errors in C files. Maybe i could run a formatter for those.

PR #26, opened 2015-06-26
drawlib optimisation and several warnings fixed

dukesrg:


dukesrg:
After fixing makefiles there is no more need to add any devkitARM into path.

rxtools/source/lib/draw.c
```diff
@@ -167,25 +164,14 @@ void DrawPixel(int x, int y, int color, int screen){
...
-        *(scr++) = *(top++);
-        if(TOP_SCREEN2){
-            *(mir++) = *(tmp++);
-            *(mir++) = *(tmp++);
-            *(mir++) = *(tmp++);
-        }
-	}
+	memcpy(TOP_SCREEN, GetFilePack(TOP_PIC), SCREEN_SIZE);
+	if(TOP_SCREEN2)
+		memcpy(TOP_SCREEN2, GetFilePack(TOP_PIC), SCREEN_SIZE);
```
Syphurith: Yes that's quite better than what i did.

PR #27, opened 2015-06-27
Update main.c - Clear Bottom Screen

ApacheThunder:
Clear bottom screen to black on boot. It's an eye sore. Especially with the new GUI. Tested this and it works! White screen was probably a side effect of spider payload. This change to main.c will clear bottom screen during initial boot of rxTools and set it to black.

If you ever decide to put a logo there, you can remove this. Otherwise this should do well in the mean time!

### DIFF
--- a/rxtools/source/main.c
+++ b/rxtools/source/main.c
@@ -14,6 +14,7 @@
 
 void Initialize()
 {
+	ClearScreen(BOT_SCREEN, BLACK);
 	DrawString(TOP_SCREEN,  " INITIALIZE... ", 0, 240-8, WHITE, BLACK);
 	if(InitFS()){
 		DrawString(TOP_SCREEN,  " LOADING...    ", 0, 240-8, WHITE, BLACK);
--- a/rxtools/source/main.c
+++ b/rxtools/source/main.c
@@ -30,6 +30,7 @@ void Initialize()
     rxMode_boot();
 	
 	rxTools_boot:
+	ClearScreen(BOT_SCREEN, BLACK);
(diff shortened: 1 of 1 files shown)


ApacheThunder:
Ok after further tests you may need to add "ClearScreen(BOT_SCREEN, BLACK);" to a few more locations. The original line added in the original pull request, then another after "rxTools boot;" (found on line 31 of original file) and again after Initialize(); but before the draw string that draws the "Support the original, not the imitation" (line 49 of original file)

Even then it will still sometimes not clear the bottom screen (that spiderhax screen white can be stubborn it seems. :P ).

Perhaps changes to draw.c to make it more reliable are needed. The other contributors here can perhaps take a look at improving it. ;)

dukesrg:
Using BOTTOM_SCREEN is unsafe now, all tor draw routines is unaware of the different botom scrren size for now. In case of ClearScreen it will definitely write some data out of the framebuffer.

Syphurith:
Just as dukesrg said. The BOTTOM_SCREEN buffer isn't well known.
Eitherway this is ok - if you've tested it already. I didn't remember where it is ever used.

roxas75:
No it's not safe. He probably tested it only from spider, but on mset and ninjhax will not work, and will probably freeze due to an unspecified pointed memory area.
For now i'll close the pull, just becouse i'll think myself about it when i'll update the whole GUI

ApacheThunder:
Ahh I see. Well it's definitely an issue that needs looked at. Clearing/setting bottom screen to black is only needed for Spider I think. It's already black for MSET/CN users.

PR #28, opened 2015-06-27
Improvements to the FBI injection procedure.

DarkMatterCore:
These improvements include:

* New code to check the console region by reading the SecureInfo_A/SecureInfo_B file.
* Changes to the way the FindApp() function handles paths.
* Compatibility with the TMDs generated by the CIA NAND Inject Generator created by RikuKH3. It is now necessary to also have the "fbi_inject.tmd" file in the SD card.
* SHA-256 hash verification (through polarssl/sha2.h), by calculating sums and comparing them with the TMD info.
* Health & Safety App TMD & Content file backup.
* New option to restore the H&S TMD & Content file backup.

H&S backup data is created in "SD:/rxTools/h&s_backup/[Region String]/v[TMD version number]".

* [Region String] can be any of the strings available in the regions[] array.
* [TMD version number] is determined by the application at runtime by reading the Title Version from the TMD.


dukesrg:
Just wondering, it there a special purpose of passing drive as a char*?
Maybe it is better to use just a char with a 0/1 values and use %d in filename format string?
Was going to fix that along with NandSwitch() return value, but will wait until @roxas75 revise your code.

DarkMatterCore:
No special purpose at all, I just didn't mind it too much. But you are right, though. I'm still interested to know Roxas75 opinion.

rxtools/source/features/downgradeapp.c
```diff
@@ -11,169 +11,582 @@
...
-    char* folder = &tmpstr;
+char cntpath[256]; // Contains the NAND content path
+char tmdpath[256]; // Contains the NAND TMD path
+
+void print_sha256(unsigned char hash[32])
+{
+	int i;
+	for (i = 0; i < 32; i++)
+	{
+		print("%02X", hash[i]); ConsoleShow();
```
Syphurith: Well maybe you don't need to flush the console this frequently.
Just to flush it when `\n` is printed or finished is ok?

rxtools/source/features/downgradeapp.c
```diff
@@ -11,169 +11,582 @@
...
+	sprintf(tmpstr, "%s:rw/sys/SecureInfo_A", drive);
+	opened = FileOpen(&secureinfo, tmpstr, 0);
+	if (!opened)
+	{
+		print("Error.\nTrying with SecureInfo_B... "); ConsoleShow();
+		sprintf(tmpstr, "%s:rw/sys/SecureInfo_B", drive);
+		opened = FileOpen(&secureinfo, tmpstr, 0);
+		if (!opened)
+		{
+			print("Error.\nProcess failed!\n"); ConsoleShow();
```
Syphurith: Why not just return the value after the ConsoleShow? Same for Line 149.

rxtools/source/features/downgradeapp.c
DarkMatterCore: I actually forgot to change that, thanks.

rxtools/source/features/downgradeapp.c
```diff
@@ -11,169 +11,582 @@
...
     unsigned int titleid_high[] = { 0x00020000, 0x00021000, 0x00022000/*, 0x00026000, 0x00027000, 0x00028000*/}; //JPN, USA, EUR, CHN, KOR, TWN
-    char* regions[] = { "Japan", "USA", "Europe", /*"China", "Korea", "Taiwan", */"Unknown"};
     unsigned int mset_hash[] = { 0x57358F14, 0xA28EAD9F, 0x530C345B };   //JPN, USA, EUR
 	unsigned int supported_regions = sizeof(titleid_high)/sizeof(titleid_high[0]);
-    unsigned int region;
-    ConsoleInit();
-    ConsoleSetTitle("MSET Downgrader");
 	
+	ConsoleInit();
+    ConsoleSetTitle("MSET Downgrader");
```
Syphurith: Well.. I think space/tab is somewhat an not important issue now..

rxtools/source/features/downgradeapp.c
```diff
@@ -11,169 +11,582 @@
...
+			if (FindApp(titleid_low, titleid_high[region], "1")) // SysNAND only
+			{
+				FileOpen(&dg, cntpath, 0);
+				unsigned int check_val;
+				FileRead(&dg, &check_val, 4, 0x130);
+				FileClose(&dg);
+				
+				if (check_val != 0)
+				{
+					if (checkDgFile(dgpath, mset_hash[region]))
```
Syphurith: If `goto LABEL` is allowed. This might be done more beautifully.

rxtools/source/features/downgradeapp.c
```diff
@@ -11,169 +11,582 @@
...
+																	
+																	/* Verify the Content Data hash */
+																	sha2((unsigned char*)(buf + 0x1000), fbi_cntsize, CntDataSum, 0);
+																	if (memcmp(CntDataSum, TmdCntDataSum, 32) == 0)
+																	{
+																		/* Now we are ready to rock 'n roll */
+																		if (FileCopy(tmdpath, "0:fbi_inject.tmd") == 1)
+																		{
+																			if (FileCopy(cntpath, "0:fbi_inject.app") == 1)
+																			{
```
Syphurith: Oh.. Well. Similar as what above..

rxtools/source/lib/fs.c
```diff
@@ -99,22 +99,48 @@ void FileClose(File *Handle)
...
+	unsigned int size = FileGetSize(&in);
+	if (size == 0)
+	{
+		FileClose(&in);
+		FileClose(&out);
+		return -1;
+	}
+	
+	int pos = 0, res = 1;
+	unsigned int i, chunk_size = 0x4000;
```
Syphurith: Well i think maybe we should use fatfs directly for this. However forget this today.

rxtools/source/features/downgradeapp.c
Syphurith: @DarkMatterCore Thanks. If you want, you can change & test & push, just before this is merged..

rxtools/source/features/downgradeapp.c
Syphurith: *Forget this please*.

rxtools/source/features/downgradeapp.c
DarkMatterCore: I'm on it. :)

rxtools/source/features/downgradeapp.c
DarkMatterCore: About this, I prefer to wait until Roxas75 decides what's best. I'll update my code if he says the use of goto jumps is ok.

Syphurith: @DarkMatterCore Okey, as that is only a suggestion.

roxas75:
There actually is no different for passing the drive as char* or as an int.
I like all the code, and i assume it is working.
Can i merge?

dukesrg:
For just passing there is no difference, right. But when it comes out that that drive is got from NabdSwitch choice, then analyzed, passed and finally sprinted. If you don't mind I'll try to rewrite all that chain into char or int

PR #32, opened 2015-06-28
Linked NAND/drive with int value

dukesrg:


rxtools/source/features/NandDumper.c
```diff
@@ -26,17 +26,17 @@ int NandSwitch(){
...
     }
 }
 
 void NandDumper(){
 	File myFile;
-	int isEmuNand = checkEmuNAND() ? NandSwitch() : 0;
-	if(isEmuNand == -1) return;
-
+	int isEmuNand = SYS_NAND;
+	if(checkEmuNAND() && (isEmuNand = NandSwitch()) == UNK_NAND) return;
```
Syphurith: It is a little confusing. Since `UNK_NAND` means Pressing B when asking for NAND partition, so what happens if there isn't EmuNAND present? `NandSwitch` would return 0, which means `SYS_NAND` some commits before, so it wouldn't return isn't it?

dukesrg: NandSwitch() invoked only if checkEmuNAND() returned 1, i.e. emuNAND present. The save behaviour as before
No emunand - proceed with sysnand
Emunand present - ask which nand to use, is dialog cancelled - abort.

Syphurith: @dukesrg I forget its function. Oh, and i suggest you to use your enum definitions.
Switch a SYS_NAND=1, minus it to 0, and compare -- really confused me a little. sorry.

PR #33, opened 2015-06-28
FS: lseek tweak, FSFileCopy, Rearranged+formatted

Syphurith:
`f_lseek` is included in `FileRead` and `FileWrite`, however this part of code just execute.. Also it's quite slow with NAND at least on STM32F2. So I use `f_tell` to check where it is.
Hope this would make FS operations much faster. Yep I've also set `f_sync` to real handle.
I've also implemented a `FSFileCopy` that uses FATFS directly - **not fully compatible** with `FileCopy`. Same arguements accepts however the return value is compound one.
I've used [CoolFormat](http://blog.csdn.net/akof1314/article/details/5355948) (Chinese) to format the code to Java-Attach Style, and let it use Tab indention, after rearranged the functions according to their purpose.

I've got some questions for discuss:
* Do you really need those FileWrite/FileRead/FileClose? These are almost only wrappers for actual FATFS calls, which could be considered as not so useful, and trimmed those return values.
* Do you need several functions, providing simple FS operations, such as directly write some data to one file, or read all contents out from one.
* `fs.c` contains both FileSystem and NAND operations. I don't know if i should seperate them.

Almost forget: **i've changed NandInitializeCrypto to FSNandInitCrypto a bit**, which is called by InitFS.
BTW, i've seen some big files in the repo. So i had a long time waiting for uploading the edits.

### DIFF
--- a/rxtools/source/lib/fs.c
+++ b/rxtools/source/lib/fs.c
@@ -51,7 +51,7 @@ bool InitFS()
 	int res = 1;
 	if(f_mount(&fs[0], "0:", 0) != FR_OK) res = 0;
 	if(f_mount(&fs[1], "1:", 0) != FR_OK) res = 0;
-	if(f_mount(&fs[2], "2:", 0) != FR_OK);
+	if(f_mount(&fs[2], "2:", 0) != FR_OK) res = 0;
 	return res;
 }
 
@@ -67,7 +67,7 @@ bool FileOpen(File *Handle, const char* path, bool truncate)
     unsigned flags = FA_READ | FA_WRITE;
     flags |= truncate ? FA_CREATE_ALWAYS : FA_OPEN_EXISTING; //: FA_OPEN_ALWAYS;
     bool ret = (f_open(Handle, path, flags) == FR_OK);
-    f_lseek(Handle, 0);
+    if (f_tell(Handle)) f_lseek(Handle, 0); //Only seek to head if not
 	f_sync(&file);
     return ret;
 }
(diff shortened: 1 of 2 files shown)


Syphurith:
@roxas75 Oh well you're so fast, that i didn't get some commits uploaded in time.
Have you ever looked at those questions in the OP?

roxas75:
Yep sorry.
I looked the code and actually liked what you did.
As for the FS functions, i found them useful and faster than the original FatFS ones.
Also, if you want we can separate the nand operations from fs.c, i've done like this just becouse it was faster.

Syphurith:
*faster*? what i mean by Q1 is that: the FileRead/FileWrite/FileClose have just similar prototypes as original FATFS ones do. I don't find any real differences except a `f_sync` or `f_lseek` call. They calls FATFS ones, so it can never be faster than the callees.
Those could be done manually, however, dealing with the original FATFS return value may make you uneasy. **FATFS returns value telling you the reason why it failed**, and that's why i prefer it.
Well still it could be acceptable just cause its simplified operations.
Also hope someone could test the FSFileCopy out. If success it just return 0, else for errors.

@dukesrg @DarkMatterCore Or anyone else would like to discuss the questions?

DarkMatterCore:
"Do you need several functions, providing simple FS operations, such as directly write some data to one file, or read all contents out from one."

Yes. In my current code, I use FileWrite() and FileRead() quite a bit instead of directly using FileCopy() (I only use it to replace the files after all the checks have been passed). This is because I need to calculate SHA-256 sums and compare them with the TMD info.

I think we can safely replace FileCopy() with FSFileCopy(), though.

Syphurith:
Yes, FileGetSize->f_size, FileClose->f_close is quite safe to be done. And minor performance changed. FSFileCopy returns `(STEP << 8)|FatfsRetval` that's why i said that's not fully compatible (when success just 0). I could think of the potential perfermance differences as answer. The questions are answered already. "Not so worthful". Thank you @DarkMatterCore.

PR #34, opened 2015-06-28
FS: split FS and NAND operations

Syphurith:
I've splitted those operations so there is two new files named as `nand.c` and `nand.h`.
I modified FSInit so that should return the correct return value. Also renames.

I was working on these changes, but didn't complete in time, orz. This should be part of #33, sorry.


roxas75:
That's it, sorry for the hurry ahaha

Syphurith:
@roxas75 Thanks. No need to say sorry for me, that's my fault don't push it in time.

PR #35, opened 2015-06-28
fill_* tools substituted by dd

dukesrg:
It is much faster and do not need to build additional tools that way.
/dev/zero may be changed to /dev/random or /dev/urandom to get crap generated :)


DarkMatterCore:
Have you tested it with a dd build for Windows/other OS?

dukesrg:
I only tested with Windows (devkitpro/msys build)

DarkMatterCore:
Cool. This should reduce the compilation problems some people were getting (including myself, I had to use a different MinGW GCC build for these tools for some reason).

dukesrg:
Well, there are several more tools which requires gcc to build for the build machine platform. So additional MinGW/cigwin/msys is required to build. However installing MinGW is very easy IMO.

PR #36, opened 2015-06-28
Comparision between signed and unsigned integer

AnsgarPfeifer:
Only a small compile error fix

### DIFF
--- a/tools/toolsrc/fill_with_zero/main.c
+++ b/tools/toolsrc/fill_with_zero/main.c
@@ -15,8 +15,7 @@ int main(int argc, char** argv){
 	unsigned int size = atoi(argv[2]);
 	printf("%d bytes\n", size);
 	fseek(pf, 0, 2);
-	unsigned int fsize = ftell(pf);
-	while(fsize < size)
+	while((unsigned) ftell(pf) < size)
 		fputc(0, pf);
 	fclose(pf);
 	return 0;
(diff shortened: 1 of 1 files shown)


tools/toolsrc/fill_with_zero/main.c
```diff
@@ -15,8 +15,9 @@ int main(int argc, char** argv){
 	unsigned int size = atoi(argv[2]);
 	printf("%d bytes\n", size);
 	fseek(pf, 0, 2);
-	while(ftell(pf) < size)
+	unsigned int fsize = ftell(pf);
+	while(fsize < size)
```
dukesrg: It will be an endless loop. ftell returns current position at the moment, not the pointer to the current position.

AnsgarPfeifer: Okay thanks, fixed

AnsgarPfeifer:
Don't know how to change more files in this pull, but same error in "tools/toolsrc/fill_with_crap/main.c".
Maybe somebody could fix this.

dukesrg:
There are no errors in that files, just get the right gcc compiler with MinGW and set oath variables.

Syphurith:
@AnsgarPfeifer Thanks as always for contributing. Ye the some *warning fix*s could be bundled into one pull, after you've understand those code. You could even add comments and descriptions.

PR #38, opened 2015-06-29
Rxmode: format, indent, rx_memmem

Syphurith:
This brings these edits: *Now went for food..*
* Formatted, indented (tabs), used u8 (those short terms).
* `rx_memmem` for search a pattern (binary is ok) within a given buffer.
* ~~Renamed `MyThread` to `rxModeThread`.~~

About the `rx_memmem`, if you want to see that's actually working, tell me and i would show you the test program on my Windows 8.1 x64. I'm sure it should be faster than the original method.

### DIFF
--- a/rxmode/native_firm/source/arm9/source/myThread/FS.h
+++ b/rxmode/native_firm/source/arm9/source/myThread/FS.h
@@ -1,9 +1,9 @@
 #ifndef FS_H
 #define FS_H
 #include <stdio.h>
-extern unsigned int fopen9(void *handle, wchar_t* name, unsigned int flag);
-extern void fwrite9(void* handle, unsigned int* bytesWritten, void* dst, unsigned int size);
-extern void fread9(void* handle, unsigned int* bytesRead, void *src, unsigned int size);
+extern unsigned int fopen9(void *handle, wchar_t *name, unsigned int flag);
+extern void fwrite9(void *handle, unsigned int *bytesWritten, void *dst, unsigned int size);
+extern void fread9(void *handle, unsigned int *bytesRead, void *src, unsigned int size);
 extern void fclose9(void *handle);
 
 #endif
--- a/rxmode/native_firm/source/arm9/source/myThread/lib.c
+++ b/rxmode/native_firm/source/arm9/source/myThread/lib.c
@@ -4,8 +4,8 @@
 // ---------------------------------------- //
 #include "lib.h"
 
-void *memset(void * ptr, int value, unsigned int num){
-    unsigned char *p = ptr;
+void *rx_memset(void * ptr, u8 value, u32 num){
+    u8 *p = ptr;
     while (num) {
         *p++ = value;
         num--;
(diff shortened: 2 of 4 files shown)


roxas75:
Can you please rebase your changes with the new rxMode setup?
Also, thanks for the changes.

Syphurith:
@roxas75 I would. Also, there is some small areas i wanna fix it more.

Syphurith:
@roxas75 oh.. could you tell me how to use the `fopen9` inside rxMode, FS.s?
I want to test if i can cache the value..

roxas75:
*unsigned int fopen9(void *handle, wchar_t* name, unsigned int flag);*
- Handle : I'm not sure how much space it takes. I think around 8 bytes, though make it like a 32 bytes array is safer. It needs to be cleared to all zeros to work.
- Name : The filepath of the file, as wide char. The main archives you can need are **sdmc:/** and **nand:/**
- Flag : They are documented on 3dbrew, it gives you the privileges to the file. The bits are : 0 = read; 1 = write; 2 = create;

It should return 0 if the file is successfully opened.

Syphurith:
@roxas75 Ok, just forget about the renamed "rxModeThread". Rebased.
I would try to cache the offset so that it could be launched faster. So please do not insta merge.

roxas75:
Did you test it?
Just becouse mset and menu patching is done by some studied timing patching, so i wonder if you changed it in the proper way.

Syphurith:
@roxas75 Now time for testing the new build :). Timing patching? Please tell me more about the actual behavior that you want it to express.

Syphurith:
Well as it seems to stop working after my changes. I think i could close it.

Syphurith:
Eh.. Could you tell me the normal first-time boot after 2.6? @roxas75 I seem to be stucked here.

Syphurith:
@roxas75 Eh.. if rebased on commits provided that tagged as 2.6 it would almost halt when "LOADING". I've compiled the source that tagged as 2.6 and that still halt. With the "release.rar" it does boot, and seems to have some issues with Linked/Unlinked NANDs, which then halts at the "Installing to SD".. Note: Linked/Unlinked issue happens with at least MSET loading.
Also i think it's time for me to close this pull request now. Since memmem is present in Linux, i think you can get it somewhere else if pleased. If you really need the implementation, tell me please.

Syphurith:
@roxas75 Haha, that's more interesting than i've ever thought. You know the system configuration inside Emunand has "RX-E" or something else right? At least i can't be just "Ver.". However, when i picked those commits to remake them before 2.6, it really boots and it showed "Ver. 9.8.0-10J" where i've just tried DevMenu. It seems its offset differs from mset is wrong.. Well amusing.
I would try to install a CIA from other regions to see if the region free works after these commits.
Orz it does boot a USA title from JPN console.. also Icon present.. Since 2.6 is already out, i think i would not push just commits before i succuessfully boot the new one. No offset caching currently.

Syphurith:
Eh.. I would eat my words this time. After reboot all out-of-region icons disappeared.

PR #40, opened 2015-06-29
Makefile fixed, insert.py substituted by dd

dukesrg:
Build tools dependencies was somehow removed in the last commit. Fixed that, now builds fine from the clear clone. Also insert.py is no more used, dd insertion is way faster.

P.S. from the last commit I noticed rxtools.dat have integrated timestamp, so diff with subsequent builds fails. Was this added intentionally?


Syphurith:
@dukesrg I don't know why there is timestamp. Maybe for debug options?
BTW there is 2.6 out. Unfortunately i can't get it to run. Roxas is busying trying to fix it.

dukesrg:
Yes, I know about 2.6, this pool request is from rebased version. However only build system is affected with this request.

dukesrg:
That was a buggy rebase, now repeated it and looks OK

Syphurith:
Indeed you could rebase it many times to let it look good, then just push to original branch with unknown edits and it would show as rebased ones.

dukesrg:
Something was wrong with this rebase - it reported OK, but actually did not see the latest master branch files and form graph did not show this branch was rebased on the master. However pool request showed right changes. Clone/rebase from the scratch helped :)

PR #47, opened 2015-06-30
Fixed types, casts, not used vars which produced warnings (result binaries unchanged!)

dukesrg:
Several casts could be avoided by changing prototypes to use void* where typed (or pointer) vars not needed, but intentionally left untouched with this commit to keep compiled binaries unchanged.

### DIFF
--- a/brahma/data/payload/source/main.c
+++ b/brahma/data/payload/source/main.c
@@ -2,7 +2,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 
-void* (*rxTools)() = 0x08000030;
+void* (*rxTools)() = (void*)0x08000030;
 
 void main()
 {
@@ -30,7 +30,7 @@ void main()
 	*((unsigned int*)0x080FFFC4) = 0x20046500;
 	*((unsigned int*)0x080FFFD8) = 0;
 
-	unsigned int* buf = 0x20400000;
+	unsigned int* buf = (void*)0x20400000;
 	unsigned int base = 0x67893421;
 	unsigned int seed = 0x12756342;
 	for(int i = 0; i < 400*1024/4; i++){
@@ -38,8 +38,8 @@ void main()
 		base += seed;
 	}
 
-	unsigned char*src = 0x20400000;
-	unsigned char*dst = 0x08000000;
+	unsigned char*src = (void*)0x20400000;
+	unsigned char*dst = (void*)0x08000000;
 	for(int i = 0; i < 320*1024; i++){
 		dst[i] = src[i];
 	}
--- a/msethax/payload/source/main.c
+++ b/msethax/payload/source/main.c
@@ -2,7 +2,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 
-void* (*rxTools)() = 0x08000030;
+void* (*rxTools)() = (void*)0x08000030;
 
 void main()
 {
@@ -10,11 +10,11 @@ void main()
 	unsigned char* fb2 = (unsigned char*)(0x201CB370);
 	unsigned char pure_color = 0x77;
 	
-	*((unsigned int*)0x080FFFC0) = fb1;
-	*((unsigned int*)0x080FFFC4) = fb2;
+	*((unsigned int*)0x080FFFC0) = (int)fb1;
+	*((unsigned int*)0x080FFFC4) = (int)fb2;
 	*((unsigned int*)0x080FFFD8) = 0;
 	
-	unsigned int* buf = 0x20400000;
+	unsigned int* buf = (void*)0x20400000;
 	unsigned int base = 0x67893421;
 	unsigned int seed = 0x12756342;
 	for(int i = 0; i < 400*1024/4; i++){
@@ -24,8 +24,8 @@ void main()
 	
 	if(buf[0] != 0xE51FF004) pure_color = 0x00;
 	
-	unsigned char*src = 0x20400000;
-	unsigned char*dst = 0x08000000;
+	unsigned char*src = (void*)0x20400000;
+	unsigned char*dst = (void*)0x08000000;
 	for(int i = 0; i < 320*1024; i++){
 		dst[i] = src[i];
 	}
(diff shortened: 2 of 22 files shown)


brahma/data/payload/source/main.c
```diff
@@ -30,7 +30,7 @@ void main()
 	*((unsigned int*)0x080FFFC4) = 0x20046500;
 	*((unsigned int*)0x080FFFD8) = 0;
 
-	unsigned int* buf = 0x20400000;
+	unsigned int* buf = (void*)0x20400000;
```
Syphurith: well if change all those `unsigned int` to `u32` and `unsigned char` to `u8`, this would be `(u8*)`. It is only a value pointer cast so it is safe. (* ```void*``` is special..*)

brahma/data/payload/source/main.c
```diff
@@ -38,8 +38,8 @@ void main()
 		base += seed;
 	}
 
-	unsigned char*src = 0x20400000;
-	unsigned char*dst = 0x08000000;
+	unsigned char*src = (void*)0x20400000;
+	unsigned char*dst = (void*)0x08000000;
```
Syphurith: similar, i would like it to be u8* or char*.

msethax/payload/source/main.c
```diff
@@ -10,11 +10,11 @@ void main()
 	unsigned char* fb2 = (unsigned char*)(0x201CB370);
 	unsigned char pure_color = 0x77;
 	
-	*((unsigned int*)0x080FFFC0) = fb1;
-	*((unsigned int*)0x080FFFC4) = fb2;
+	*((unsigned int*)0x080FFFC0) = (int)fb1;
+	*((unsigned int*)0x080FFFC4) = (int)fb2;
```
Syphurith: or unsigned and signed pointers/value would cause warnings sometime.

rxmode/native_firm/source/arm9/source/myThread/myThread.c
```diff
@@ -24,17 +24,17 @@ void memdump(wchar_t* filename, unsigned char* buf, unsigned int size){
...
 
 void patch_processes(){
-	unsigned char* mset = 0x24000000;
-	unsigned char* menu = 0x26A00000;
+	char* mset = (void*)0x24000000;
+	char* menu = (void*)0x26A00000;
 	for(int i = 0; i < 0x600000; i+=4){
 		//System Menu code, which locks the region
-		if(dest == 0x20000400){	//This means we haven't still found our code
+		if(dest == (void*)0x20000400){	//This means we haven't still found our code
```
Syphurith: I don't know if compiler would complain about these casts from void* to others. To my understanding it accepts void* be assigned by others, maybe not..

msethax/payload/source/main.c
dukesrg: Not in the assignment I believe :)
Anyway we'll switch to u32 someday

rxtools/source/MainMenu.h
```diff
@@ -28,11 +28,11 @@ static void ShutDown(){
...
-		" Decrypt CTR Titles", &CTRDecryptor,
-		" Decrypt Title Keys", &DecryptTitleKeys,
-		" Generate Xorpads", &PadGen,
-		" Decrypt partitions", &DumpNandPartitions,
-		" Generate fat16 Xorpad", &GenerateNandXorpads,
+		{" Decrypt CTR Titles", &CTRDecryptor},
+		{" Decrypt Title Keys", &DecryptTitleKeys},
+		{" Generate Xorpads", &PadGen},
+		{" Decrypt partitions", &DumpNandPartitions},
+		{" Generate fat16 Xorpad", &GenerateNandXorpads},
```
Syphurith: Yes, that's easy to get rid of some warnings related to structure.

rxtools/source/features/CTRDecryptor.c
```diff
@@ -8,13 +8,13 @@
 #include "hid.h"
 #include "ncch.h"
 #include "crypto.h"
+#include "stdio.h"
 
-#define BUFFER_ADDR ((volatile u8*)0x21000000)
+#define BUFFER_ADDR ((u8*)0x21000000)
```
Syphurith: Be careful when using u8* not vu8* that contains volatile attribute. Someone compiler would like to cache value (in macro areas). volatile tells it this should never be cached.

brahma/data/payload/source/main.c
dukesrg: I suppose several lower level libraries will use void** anyway, so in case of (u8**) we can't avoid additional type casts. While void** itself is more flexible, though less safe

rxtools/source/features/TitleKeyDecrypt.c
```diff
@@ -112,7 +112,7 @@ int GetTitleKey(unsigned char* TitleKey, unsigned int low, unsigned int high){
     unsigned int tid_high = ((high>>24)&0xff) | ((high<<8)&0xff0000) | ((high>>8)&0xff00) | ((high<<24)&0xff000000);
     unsigned int tick_size = 0x200;     //Chunk size
     if(FileOpen(&tick, "1:dbs/ticket.db", 0)){
-        unsigned char* buf = 0x22000000;
+        unsigned char* buf = TITLES;
```
Syphurith: Even myself didn't see that yesterday. Using macro to expand its cast is quite good.

Syphurith:
Thank you very much for taking look at these. Most of them looks good to me. However I don't know if you've tested it yet, and if there is any conflicts with the current master.

msethax/payload/source/main.c
Syphurith: @dukesrg Okey.. Still the code would oneday be looked throughout.

rxmode/native_firm/source/arm9/source/myThread/myThread.c
dukesrg: Compiler don't complain about void* casts, all checks left on a programmer. Don't forget this actually do not alter binary, it's just don't generate warning

brahma/data/payload/source/main.c
Syphurith: @dukesrg Yeh. Have you ever tested your new build yet?

rxmode/native_firm/source/arm9/source/myThread/myThread.c
Syphurith: Good if that wouldn't generates warning like when i tried to cast malloc returned pointer to char*.

PR #51, opened 2015-07-01
SVC 0x26 AGB_FIRM patch + Code fixes + Console output cleanup (downgradeapp.c)

DarkMatterCore:
ARMIPS doesn't support the SVC (0xEF) opcode, so I had to implement the patch as a hardcoded solution. Basically, it patches the AGB_FIRM at 0xD7A10 (0x805E010 in Process9 memory area) to use the SVC 0x26 interrupt instead of 0x01.

BTW, I don't know if you guys have any problems with making the "rxTools/data" directory visible. Users already know about its existence, so I don't really think we need to hide it anymore.

**EDIT:** Here's the full changelog:

* SVC 0x26 AGB_FIRM patch (GBA BIOS bootscreen) can now be enabled through the Settings menu.
* Settings code modified to make sure unhandled exceptions won't take place.
* Pressing left or right while selecting a theme will now jump to the next available theme directory.
* InstallConfigData() now properly waits for any key to be pressed instead of the A button only.
* The application now displays the R button message in the bottom screen during the autoboot sequence.
* Fixed a small bug in FindApp() where the TMD path would be lost if another TMD with no content file was opened.
* Console output cleanup in downgradeapp.c.
* InstallData() now properly closes the firmware.bin file in case an error occurs.


motezazer:
Does everyone want to see the bootscreen?
It would be better to have it as an optional feature.

DarkMatterCore:
Maybe we can add it as a Makefile option.

dukesrg:
@DarkMatterCore 
Do you have GAS -compatible source? It it not  aproblem to add it, I can alter make files for that patch with gnu asm.
Alternatively tou can add SVC EF ARM opcode like .word 0xEF0000EF

DarkMatterCore:
@dukesrg I'll look into that.

Do you guys prefer to always ask the user on rxMode boot if they want to enable the AGB_FIRM BIOS? Or to simply add this as a build option? Because I already have added both ways, but I want your opinion.

dukesrg:
@DarkMatterCore for building patch from as source file just have a look here
https://github.com/dukesrg/rxTools/blob/0d50c4d0c0caef3e460b62f6db4c89a838be4598/rxmode/native_firm/MakefileL27
For using SVC from armips source:
https://github.com/roxas75/rxTools/blob/master/rxmode/native_firm/source/p9_reboothax.sL71
Just try to avoid including precompiled patches it it can really be made with the current toolchain

DarkMatterCore:
That should take care of things. Thanks @dukesrg.

DarkMatterCore:
That was actually my bad, I still have a long way to go in ASM. I did read about SWI, but its opcode is different, so I didn't know how to properly implement the SVC call and decided to hardcode it after testing it. Thanks for pointing it out, though.

Syphurith:
@DarkMatterCore Or else if that doesn't work, you can try the noarmips that @dukesrg holds later. I don't know much about ARM ASM/ABI either (Well even Thumb). Eh Hope [This](infocenter.arm.com/help/topic/com.arm.doc.dui0203j/Cacdfeci.html) could help you.
The handler for SVC could be implemented in C/ASM however not for us. You might need to replace the SVC value/vector if you need to replace an original handler for a certain SVC call. You could find out such things inside NTRDisasm project with IDC loaded to IDB where 44670 tried to replace a call. Once you need the original call offset, you could embed it into your new code.
And for those SVC prototypes, see those from [3dbrew](http://3dbrew.org/wiki/SVC).
Is this means you want to patch offset 0805E010 to let it be "SVC 0x26"?

DarkMatterCore:
It already works, I have tested each commit before pushing the changes to my repo. The solution proposed by @dukesrg proved to be useful.

But this can also be added as a configurable option (because, like @motetazer said, not everyone may want to see the GBA bootscreen). What do you think about this matter? @Syphurith

Syphurith:
@DarkMatterCore It does, we actually need some configuration UI. So that we could get rid of those we don't care - as for myself i don't need to play GBA ones on my O3DS.
Also, @AlbertoSONIC I think we could seperate those helper Apps for user into another repo. This one is already complex. BTW, do you need a "Buffer ^0xFF" so to highlight some area?

motezazer:
I agree with @Syphurith : we need a configuration UI.
So I close this pull request.
You are free to propose it again when the new GUI will be done.

Syphurith:
@motezazer Eh i hope most of those configuration UI could be done in rxTools as Menus, not as Apps running on PC. I've ever got a menu prototype when i was feeling bad with its menu:
```C
typedef struct {
  char* Name;
  void* SubMenu;
  NULLFUNC Call; 
} MenuItem;
typedef struct {
  char* Name;
  char Selected;
  char ItemCount;
  MenuItem* Items;
} Menu;
```
However that is for text menu, and i've ever made that working. It would need much modification to let code adjusts with the definition. It could be easily adjusted to a UI menu..
Also, you could just leave this open next time, so when it is ready to be merge just rebase and ok.

Also for @nastys I hope you please not add those Apps directly into this repo - just a suggestion. You could seperate it into another repo if you really need it badly for rxtools.

motezazer:
@Syphurith If you look at our latest Pasta build, there is an in-app settings screen, no need for the companion app anymore. You can use both.

I reopen this pull request. I closed it because I'm not in the GUI development, but it should be easy to merge.

Syphurith:
@motezazer Yes thanks for selecting the in-app UI. Also, this is for rxmode, as you can see where the files go. It isn't related to UI much, so safe till anyone else change rxMode much.
BTW i do wanna know what would be added/ported from Pasta UI. Do you need Buffer XOR?

motezazer:
@Syphurith The UI is not my domain, ask @AlbertoSONIC.

Syphurith:
@motezazer @AlbertoSONIC XOR is for fast reverse-able color change, which doesn't require more memory. See FF0000 >(XOR FFFFFF)> 00FFFF. Easy to be done, a way for highlight an area.

DarkMatterCore:
We could just save configuration info to "rxTools/data/data.bin". The application already generates this file on the first boot, and only uses it to store the build date and time. Why don't we use it for this, too?

I can add a quick config menu that saves the options using a simple struct.

motezazer:
@DarkMatterCore Wait for the new GUI that @AlbertoSONIC develop.

DarkMatterCore:
Just updated my rxTools fork. The latest commit adds the possibility to check the EmuNAND (if available) and get the AGB_FIRM Title Key from there in case it isn't installed in SysNAND. This should probably reduce most AGB_FIRM-related installation problems during the initial setup.

@AlbertoSONIC Any plans on collaborating with Crusard to make an in-app settings screen?

motezazer:
@DarkMatterCore There is an in-app settings screen!

DarkMatterCore:
@motezazer Oh, I didn't notice it. Only the "Press A" screen is available at this moment, though.

AlbertoSONIC:
There's a full settings app which allows to enable the boot gui everytime and to select a theme!

DarkMatterCore:
It is text-based, after all. Thanks. I'll add the AGB_FIRM BIOS option as soon as possible.

DarkMatterCore:
It should be ready now, please check it. I took the liberty to modify the Settings code a bit to make sure unhandled exceptions wouldn't take place.

AlbertoSONIC:
I see you removed the check in theme selection which controls if theme++ exists before making it available in settings.. You have to re-add it, otherwise if an user selects theme "5" and it doesn't exist, it will not load any theme until next boot

DarkMatterCore:
But if a theme "6" exists, the user won't be able to select it unless the theme directory is renamed. I'll see what can I do, though.

DarkMatterCore:
@AlbertoSONIC It should be OK now.

DarkMatterCore:
Fork updated again. I have modified the first comment to add the complete list of changes.

AlbertoSONIC:
I guess I can merge, right? I can't test it a atm...

DarkMatterCore:
I have already tested each feature. It works fine. ;)

PR #54, opened 2015-07-02
Draw lib cleanup

dukesrg:
Get/DrawPixel prototype fixed with pointer type
Functions parameters switched to types defined in common.h
Initial support for soft alpha channel

### DIFF
--- a/rxtools/source/lib/draw.c
+++ b/rxtools/source/lib/draw.c
@@ -7,165 +7,155 @@
 #include "draw.h"
 #include "filepack.h"
 
-int current_y = 1;
+u32 current_y = 1;
 
-u8* tmpscreen = (u8*)0x26000000;
+u8 *tmpscreen = (u8*)0x26000000;
 
-void ClearScreen(unsigned char *screen, int color)
+void ClearScreen(u8 *screen, u32 color)
 {
-    u32 i = SCREEN_SIZE/4;  //Surely this is an interger.
-    u32* tmpscr = (u32*)screen; //To avoid using array index which would decrease speed.
-    //Prepared 3 u32, that includes 4 24-bits color, cached. 4x(BGR)
-    u32 color0 = (color) | (color << 24),
-        color1 = (color << 16) | (color >> 8),
-        color2 = (color >> 16) | (color << 8);
-    while (i--) {
-        *(tmpscr++) = color0;
-        *(tmpscr++) = color1;
-        *(tmpscr++) = color2;
-    }
+	u32 i = SCREEN_SIZE/sizeof(u32);  //Surely this is an interger.
+	u32* tmpscr = (u32*)screen; //To avoid using array index which would decrease speed.
+	//Prepared 3 u32, that includes 4 24-bits color, cached. 4x(BGR)
+	u32 color0 = (color) | (color << 24),
+		color1 = (color << 16) | (color >> 8),
+		color2 = (color >> 16) | (color << 8);
+	while (i--) {
+		*(tmpscr++) = color0;
+		*(tmpscr++) = color1;
+		*(tmpscr++) = color2;
+	}
 }
 
-void DrawCharacter(unsigned char *screen, int character, int x, int y, int color, int bgcolor)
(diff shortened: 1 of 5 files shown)


rxtools/source/lib/draw.c
```diff
@@ -7,165 +7,155 @@
...
-    //Prepared 3 u32, that includes 4 24-bits color, cached. 4x(BGR)
-    u32 color0 = (color) | (color << 24),
-        color1 = (color << 16) | (color >> 8),
-        color2 = (color >> 16) | (color << 8);
-    while (i--) {
-        *(tmpscr++) = color0;
-        *(tmpscr++) = color1;
-        *(tmpscr++) = color2;
-    }
+	u32 i = SCREEN_SIZE/sizeof(u32);  //Surely this is an interger.
```
Syphurith: I'm sorry but we may need to change it to `SCREEN_SIZE/BYTES_PER_PIXEL` cause it should means WIDTH*HEIGHT. Or else it would forget some pixels.

rxtools/source/lib/draw.c
```diff
@@ -7,165 +7,155 @@
...
-    vasprintf(&str, format, va);
-    va_end(va);
+	va_start(va, format);
+	vasprintf(&str, format, va);
+	va_end(va);
 
-    DrawString(TOP_SCREEN, str, 10, current_y, RGB(255, 255, 255), RGB(0, 0, 0));
-    free(str);
+	DrawString(TOP_SCREEN, str, 10, current_y, RGB(255, 255, 255), RGB(0, 0, 0));
+	free(str);
```
Syphurith: Even this function isn't called, free a already defined array isn't safe.

rxtools/source/main.c
```diff
@@ -42,15 +42,15 @@ void Initialize(){
 		u32 pad = GetInput();
 		if(pad & BUTTON_R1 && i > 0x333333) goto rxTools_boot;
 	}
-    rxModeQuickBoot();
+	rxModeQuickBoot();
```
Syphurith: I suggest to make some text before quickboot, also some after it.
So when it failed you can be able to know when.

Syphurith:
BTW, @dukesrg I think we may make those Ram Obsolute Addresses that appeared in the files to be defined when booting entrypoints. Thus you may need to define a small series of pointer as structure:
```C
#ifndef BOOT_ADDR //If this address store isn't defined in Makefile
#define BOOT_ADDR (u8*) 0x20040000 // Warning: only a demo.
#endif
typedef struct {
  void** VRAM0; //Be void**, cause it holds address to actual buffer (void*)
  void** VRAM1;
  void** BUF0;
....//Surely you could change it, this is only a demo
  u8 TYPE; //Telling us what kind of entrypoint we've passed
} bootAddr;
#define BOOT_ADDRESS (bootAddr*)BOOT_ADDR //Even used only once.
....
void* VRAM = 0; //all those global.
void CfwAddrInit(void) { //And those values are passed here.
  VRAM = BOOT_ADDRESS->VRAM0;
}
...
int main(void) {
  CfwAddrInit(); //Init it, be the first.
...
}
```
So we could deal with more kinds of entrypoints.
@mid-kid @motezazer What do you think of this?

rxtools/source/lib/draw.c
dukesrg: Screensize is already w*h*bpp
/sizeof(u32) means we iterate by 32bit increment

Syphurith: @dukesrg Thanks. My mistake doesn't make troubles. I was just waken up. Sorry being careless.

dukesrg:
@Syphurith Can't we just relocate VRAM base address to the desired location or just reset it to default system app screen buffer location, listed on 3dbrew memory map page?

P.S. bad new, my 3DS gonna die soon - got several black and blue screen boots yesterday Fixed by sqeezing 3ds :(

Syphurith:
@dukesrg WTF? Have your NAND backed up? Hope it could be repaired easily.

PR #64, opened 2015-07-05
Use $(MAKE) instead of make in Makefiles

173210:
$(MAKE) has many advantages to just calling make.
See this page for the detail.
https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,10 @@ distclean: clean
 
 .PHONY: clean
 clean:
-	@make -C rxtools clean
-	@make -C rxmode clean
-	@make -C brahma clean
-	@make -C msethax clean
+	@$(MAKE) -C rxtools clean
+	@$(MAKE) -C rxmode clean
+	@$(MAKE) -C brahma clean
+	@$(MAKE) -C msethax clean
 	@rm -f $(tools) payload.bin data.bin rxTools.dat
 
 .PHONY: release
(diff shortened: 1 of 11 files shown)


Syphurith:
Would wait to see if anyone faces problems with `$(MAKE)`. Iirc it was changed some commits ago.

173210:
Yes, I'm facing the problem. I'm aiming to use job option to make build faster to provide nightly builds.

dukesrg:
Last time I was changing $(MAKE) with make was when sources path haveva space sign. Not a big deal but newbies will be yelling about not compiling. And I don't know how to escape spaces in makefile

173210:
```
make[1]: Entering directory `/c/Users/root3/Documents/GitHub/rx Tools/rxtools'
make[2]: /c/Users/root3/Documents/GitHub/rx: No such file or directory
```
Actually it outputs errors.

Syphurith:
@dukesrg Try quote them with `"`. As in windows, many programs recognize it.

173210:
@Syphurith Exactly it's the usual way to escape spaces. However, we can't do that for $(MAKE) because $(MAKE) includes both of the command and options.
So we must evaluate which is more important, options or space escaping. I don't prefer to use bare `make` because it's not recommended by GNU.

Syphurith:
@173210 or simply tell users there should be no space when building it. Or fix it themselves?

173210:
Yes, I think it's the best way to tell users so.

dukesrg:
So we need to update Readme.md to incidate source path restriction. Because I've already troubleshooted at least one with spaces!

173210:
I close this pull request for now.

PR #65, opened 2015-07-05
Font changed to thin Eurostile-like

dukesrg:
New font as requested by @AlbertoSONIC


dukesrg:
Font with tool added in new mergeable pool request

PR #67, opened 2015-07-05
Theme generation moved to makefile

dukesrg:


Syphurith:
Eh.. Why that is rotate 90? Sorry i don't know what exactly we are trying to do with those pictures. @AlbertoSONIC If please tell me about it? I would think of using nodejs scripts to do so if possible.

nastys:
@dukesrg Adding theme generation to makefile is a good idea, but the scripts should not have been removed, because they are still useful for custom theme creators and testing. I'm going to improve the scripts and put them into a separate folder.

dukesrg:
@nastys There was only one actual convert call per file in the old scripts, I had no idea how they could be used for anything else. You can always run 'make theme' or 'cd theme; make' to just convert theme files

nastys:
@dukesrg Yes, but I assume theme creators don't have or don't know how to use "make" and also there will be a lot more scripts.

dukesrg:
@nastys For theme creators it is better to make a separate repo with all tools.

nastys:
@dukesrg My goal is to make theme creation as simple as possible. Maybe an all-in-one script would be a good compromise.

PR #69, opened 2015-07-06
CakeHax - Rebased edition.

mid-kid:
I don't know what's up with the double commits, but it works, so hey.

Anyway, I just rebased the pull request from #55 for easy merging.
There's some things, however, that still don't work: For one we have rxinstaller.nds, of which the ROP chain has to be adapted for CakeHax.
Then, we have the draw.h, which I have no idea what you want to do with it, but that's up to you (for now define BOT_SCREEN as BOT_SCREEN2).
About apache's changes, some of them (like decrypting filepack) can be done in a separate pull request. Others, I'll cherry-pick.

There's still some testing to testing to be done before this is finished.


mid-kid:
Fun fact: It doesn't get past the Loading... screen.

mid-kid:
Nevermind, I didn't have the theme installed. You should consider some proper error messages.

AlbertoSONIC:
So this is automatically mergeable?

motezazer:
Yex, it's based on the latest commit.

AlbertoSONIC:
Then we can merge it!

AlbertoSONIC:
I'll fix ninjhax BOT_SCREEN once i get home

motezazer:
Tested with Ninjhax : works perfectly well!

motezazer:
@AlbertoSONIC Everything works, no need to fix anything.

AlbertoSONIC:
Now it's perfect ;)

dukesrg:
How about rxinstaller, is it working now?

motezazer:
@dukesrg No. And I'm not able to fix it.

mid-kid:
You could use CakesROP instead, but I badly recommend against it, since it's a bunch of copy/paste work in a "just werks™" state.

dukesrg:
@motezazer I'm asking because I'm holding one pool request ftom @173210 int my noarmips branch that allows to transfer all code ftom armips to GAS without exlicit data workaround. The only thing stops me from merging it now and making a pool request intro the main thread is that rxinstalled builds something different and I'm just not able to check if it is working like that. Firm patches compiles the same (well'' except gas generated a different branch opcode in reboothax, but this can be easily tested with SSB)

PR #73, opened 2015-07-06
Transfer armips code to gas

dukesrg:
Checked on O9.2/9.8E
+spiderhax
+ninjhax
+reboothax (binary patch data changed as gas bug(?) workaround)
?mset (can't check)
?rxinstaller.nds (can't check. .nds binary differs, though mset*.bin files are the same)

P.S. still no splashscreen on start :(


173210:
:+1:

dukesrg:
@173210 any idea about mset offset calculation changed? :)

dukesrg:
Will issue a rebased one

PR #76, opened 2015-07-08
Fixed a broken function, minor tidy ups.

lavanoid:
Fixed the strip PNG function (it had "goto:EOF", at the start of it?),
tidied up the code to save a few bytes, added more detailed error
outputs etc.

### DIFF
--- a/tools/themetool.bat
+++ b/tools/themetool.bat
@@ -1,11 +1,10 @@
 @echo off
 
 echo.
-echo rxTools AIO Theme Tool v1.0.1
+echo rxTools AIO Theme Tool v1.0.2
 
 convert | findstr /C:"Miscellaneous" 1>nul
 if errorlevel 1 (
-	color c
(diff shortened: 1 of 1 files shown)


tools/themetool.bat
```diff
@@ -61,58 +61,66 @@ set /a i=%i%+1
...
+)
+if not exist "cfg0TOP.bin" (
+	echo Cannot find "cfg0TOP.bin"! Aborted.
+	goto:EOF
+)
 ren TOP.bin TOP.bi_
 ren cfg0TOP.bin cfg0TOP.bi_
-for /r %%f in (*.bin) do (
-call :topng %%f 320
+for /r %%f in (*.bin) do call :topng %%f 320
```
nastys: Why search for TOP.bi_ and cfg0TOP.bin_?
Also, I think it should continue even without hero-new-3ds.png.
Well, I'll merge it for now...

PR #84, opened 2015-07-19
Fixed Seed Crypto

mixups:
Fixed the issue in Seed Crypto not uses 7.x NCCH crypto.
For Pony Pets 3D (Japan).


motezazer:
Fixed with latest commits.

Issue #63, opened 2015-07-05
GPL


173210:
README.md says "Support the original, not the imitation!" Licensing under a copyleft license allows to declare this more clearly.
If others made a modification of this tool, we can request the source code and include their changes.

Syphurith:
If pleased it could be GPL since it's based on Decrypt9, which is stated as GPL.
Also, note that Decrypt9 has changed its sdmmc function implementations from Assembly to C.

173210:
Then it __should__ be licensed under GPL. That's copyleft. I suggest to obtain consensuses of maintainers.

Syphurith:
If that could be changed to GPL, we may have sdmmc.s changed to sdmmc.c from decrypt9.
Hope that doesn't break anything. And, GPLv2/GPLv3, or simply follow decrypt9's?

173210:
Both of them are OK. But I suggest to use GPLv3.

Syphurith:
I mean the `sdmmc.s`. That is disassembly fragment from Somwhere i don't know (maybe from GW?). So Normatt rewrote it to C implementation that wouldn't violate their own license anymore.

173210:
It's possible to include his implementation in GPLv2/v3.

idgrepthat:
lol

173210:
Suggested to the PASTA team.

dukesrg:
@roxas75, @AlbertoSONIC: Please consider this topic

173210:
PASTA team should be considering now. Let's wait for them.

AlbertoSONIC:
Waiting for @motezazer and @capito27 ;)

motezazer:
We are both OK with it.
GPLv2 was accepted by the whole team.

173210:
I'm just asking, but why not GPLv3?

motezazer:
@173210 Because it requires to give any key needed to build the program, and the N3DS keys belong to Nintendo.

173210:
@motezazer Your software is not encrypted the key and, (in theory,) it's not necessary to run rxTools.
You know, even if en/decryptor software doesn't have key, it's completely OK, of course. Keys and Nintendo firmware are just data input to rxTools.

motezazer:
@173210  New 3DS patches will need to be encrypted with the key. And we don't even know the scrambler algorithm. (So 3DS based encryption is needed)

173210:
[...]

PR #97, opened 2015-07-29
Unicode support

dukesrg:
Using GNU UniFont as a font source
Added widechar/Unicode support
Moved language-specific constants to separate header
Warnings fixed
Added unicode button characters for A/B/X/Y
Added russian translation

Limitation and bugs:
Only first 0x600 glyphs because of file size limitation (need to move font data outside) i.e. no CJK yet
A/B/X/Y button glyphs are trashed because of the above
Fixed to theme '0' because of the theme number bug appeared after switching to unicode
Several submenus with console/text output do not work/hangs
Strange va_list behaviour with %s\%S formats

upd: strange format behaviour because I was referred to MSDN format string specification instead of C99, fixing it now...

upd2: fixed all, added build-in ASCII 7-bit + external full unicode font support. Need to test this version before merging

upd3: menus and font load working fine, fixing small glitches with semi-proportional font symbols


dukesrg:
Now can be merged, fully functional with several glitches:
- escape character appeared at first position of menu console for special files dump, after that it persists in settings menu and other console menu options
- second part of nand dump progress bar have a changed character (by -1) while source have all the same

Syphurith:
@dukesrg So this clears the way to external language supports, and more customizable theme.
If that already good enough, you could change the title, just the `Edit` button around the pull name.

dukesrg:
@Syphurith Thanks, I thought I already did it yeterday :) Anyway @AlbertoSONIC  is PMed about that :)
I believe @173210 will add JSON support soon so we can move all language definitions outside

173210:
JSON implementation is already completed. I'm waiting for #63 ...

dukesrg:
@173210 I hope UniFont GPLv2+ wont' interfere?

173210:
@dukesrg It's completely OK.

PR #91, opened 2015-07-23
6.x MSET KOR Update

Gadorach:
Added 6.x KOR options to msetdg.py and downgradeapp.c

PR #103, opened 2015-08-02
6.x MSET KOR Update (mergeable)

dukesrg:
Replaces PR #91. Urgent for Korean 2DS


Gadorach:
Ah, you forgot to replace the following console printed text with "6.x MSET" instead of "5.x/6.x MSET". No worries, I have a few things I need to add to the downgrade app anyway, so I'll fix it then.

dukesrg:
If it not in Python, don't even change, all rxTools strings are moving to translation files and I'll revise it all anyway.

Gadorach:
Alright then, if you're already in the middle of setting up translatable strings, I'll wait and implement my changes after that.

PR #105, opened 2015-08-02
JSON separate language files support (don't merge yet!)

dukesrg:
- converted JSON lib to use wchar_t, UTF-8 encoding used for config file (not actually alter previous)
- added JSON templates with current file translations to /release/rxTools/lang

Known bugs:
- readCfg don't read actual config
- initial writeCfg hangs  ("Loading..." forever), while in menu write succeded, UTF-8 characters checked fine.


173210:
It's not a good idea to handle the whole JSON file with wchar_t. :-1:
You can just add CFG_TYPE_WSTRING.

soarqin:
Processing json in unicode is really a bad idea(I have not seen any json parsing library writting in unicode style)
You should have not use wchar_t anywhere (it is microsoft's dumb-ass idea which I hate very much)
It is suggested to use utf8 as 'uchar_t'(or just 'char') in C programming
For rxTools, I thinks wchar_t is only used to print unicode characters, it is better to only convert utf8 chars to wchar_t in DrawString function.

dukesrg:
@173210, @soarqin guys, dou you realize what actually is UTF8 encoding? Why the hell we should juggle bits every time we have to print a character in faver to save several bytes in Latin string while we have a plenty of ram? Saving space on disk and make easy editable - OK. But keeping it in ram - noway.

soarqin:
@dukesrg

Yes, wchar_t is ANSI part, but have you seen any famous ANSI C library using it? Have you seen any linux/BSD core projects using it? Why ANSI import wchar_t? I suggest you to check reasons, then you will know why I said it's microsoft's(and its partners) idea.

Also I think you don't know why there are both UTF-8 and Unicode(UCS-2) standard. I have been researched in character encodings for more than 10 years, as I participated in some game hackings for Chinese translations of console games several years ago, I even looked into histories of encodings.

Writing codes in UCS-2 is totally a crazy thing when you always use third-party open-souce libraries in projects or write cross-platform codes, coz most of open-source libraries doesn't support UCS-2 in their interfaces, and most of linux/BSD/Darwin syscalls/corelibs does not support UCS-2. So as a GPL project, I think we should neither use any UCS-2 in interfaces.

For characters' rendering, the most famous library is freetype. If you check freetype's render engine, you will find it does not event support unicode filenames, it only support render single characters with its UCS-2 or UCS-4 code(just only because of TTF facenames are ordered in UCS numbers), which means GTK is also render characters by converting UTF-8 to UCS-2 and store strings with char type in UTF-8. You can check their source codes.

dukesrg:
@sorqin sorry but the only reason I heard from you - "just because they didn't"
So, explain me please, why should we convert from utf-8 every time we output character if we have 64MB of free FCRAM and can make it only once? Yes, we can have UTF-8 - BMP lookup table, but again it must be done for every character for the only reason of not spending several more KB of memory for UCS-4 by fact.

173210:
```
it must be done for every character for the only reason of not spending several more KB of memory for UCS-4 by fact.
```
What do you mean with this?

soarqin:
we dont need a utf8 to ucs table, we  canjust convert it real time. if it is cpu consuming, gtk would have not using such a rendering process. So I think converting is not a big deal for cpu, it is more important to follow most of other projects coz they have mature thinks and might have some things fucked up with wchar_t, we had bettter prevent meeting unknown bugs tho.

dukesrg:
@173210 whe have a plain region of UCS2 font by fact. It will be so with any Unicode font, if we don't need UCS4. So each character may have glyph number from 0 to 0xFFFF, (except surrogates). Now we have UTF8 text which characters is encoded 1 to 3 bytes. ANSU region glyphs will be addressed directly. Anything over it will require shrinking 2 or 3 bytes into 1 or 2 bytes. I.e. 1x or 2x (shr, and, or) respectively. Instead of doing that only one on load (just because UTF8 encoded file is convinient to use) and then address any UCS2/UCS4 glyph directly.

173210:
@dukesrg It doesn't make sense because JSON won't be printed.

dukesrg:
@173210 json will also be used for external translation files!

173210:
@dukesrg So I suggest to add CFG_TYPE_WSTRINGS to convert only the value. We don't have to convert the entire file.

soarqin:
@173210 proves the thing. only printed text are needed to covert to unicode, we dont have to change json library, we k just convert language texts after they are red from json, this is better than my idea

dukesrg:
@173210 well that's a bad approach because formally all file considered utf8. Though this might work. But we will have to convert all Unicode strings separately on every language change

173210:
@dukesrg It's also true even if you converted the entire file.

soarqin:
we can do some tricks, like converting on first use

dukesrg:
Yes, we can convert the text. Having working malloc will allow to free memory occupied by miltibyte strings in json (not the case with current JSON parser implementation if we still need the rest)
Now let's count. We have additional ~9 bytes per string for quotes, colon, comma and newline, that leads to 9 additional compare value and increment pointer operation.
Say we have 100 strings, that will be 900+4(for file curls and newline) additional compare-and-incremet for converting all file at once.
Or 100 calls for mbstowcs with separate strings.

soarqin:
100 conversion of 10bytes string takes almost the same cpu time as 1 conversion of 1k, just if use -O3 switch or set inline to the function

173210:
100 calls for mbstowcs > the entire conversion

dukesrg:
@soarqin plausible, but don't forget, with separate conversions we must make it every time string is needed or occupy space with both parser multibyte string and wchar_t values.

soarqin:
@dukesrg  so I said coverting on first use

dukesrg:
@soarqin  = spreading load time on runtime while using more memory (unless we have working JSON with malloc chunks)

173210:
@dukesrg I fixed the symbol. I don't know why you think it consumes more memory.

dukesrg:
@173210 current JSON uses single character array as a source and only stores token indexes. So whenever we extract UTF-8 string from it we will be still keeping the source JSON array for any future use. So we either convert mbstowcs every time we use the string or we're to store both original json AND the converted one.

173210:
@dukesrg The current `readCfg` throw the buffer away when it ends. What is "any future use"?

dukesrg:
@173210 We will have roughly 50+ strings in translation files. And I'd rather prefer to convert it once on load and reuse token substrings directly on drawstrings.

173210:
@dukesrg Then keep only the values.

dukesrg:
@173210 Considerable. So you agree that single conversion is not so bad idea if the need of parser altering is not taken into account?

173210:
@dukesrg What is "single conversion"?

dukesrg:
@173210
fread -> mbctowcs -> JSON parse -> wcscpy/swprintf
rather then
fread -> JSON parse -> wcscpy/swprintf -> mbstowcs

soarqin:
@dukesrg just overwrite utf8 string with unicode on first use, which would not consume more memory than other approach, for example:
On loading:
```
wchar_t str[100];
((char*)str)[0] = 0;
strcpy(str+1, jsonstr);
```
In DrawString:
```
if(((char*)str)[0] == 0) {
     wchar_t tempstr[100];
     convert((char*)str, tempstr);
     wcscpy(str, tempstr);
}
DrawIt(str, ...);
```

173210:
@dukesrg I can't understand this.
```
wcscpy/swprintf -> mbstowcs
```

soarqin:
@dukesrg another problem, we are using jsmn as json parser, we had better not change its source codes at least interfaces so that we can follow its futre updates easily.

dukesrg:
@173210  sorry, I meant strcpy/sprintf -> mbstowcs - i.e. all is donw with multibyte, then only utf8 tokens encoded to ucs2.

@soarqin that will be memory hell, wchar_t string may consume more that UTF-8

soarqin:
@dukesrg unicode string length won't more 2 times more than utf8 one, so alloc twice memory of utf8 string length would be enought for later use

173210:
@dukesrg Why do you have to use strcpy/sprintf?

dukesrg:
@173210 got you, just mbstowcs :+1: Anyway I meant whole text convertion vs. only values.
@soarqin arm wchar_t is 32-bit. So for ANSI-only characters use 4 times more than UTF-8 encoded.

173210:
@dukesrg 
fread -> mbctowcs -> JSON parse -> wcscpy/swprintf
rather then
fread -> JSON parse -> mbstowcs

Can you say so?

soarqin:
@dukesrg you can use char16_t instead in C99

dukesrg:
@173210 I'd better say
fread -> mbctowcs -> JSON parse -> (wcscpy/swprintf * number_of_strings)
rather then
fread -> JSON parse -> (mbstowcs * number_of_strings)
@soarqin char16_t is C++11 UTF-16, isn't it? And substitute all swprintf whish what?

173210:
@dukesrg Why do you think so?

dukesrg:
@173210 Because mbstowcs is slower than wcscpy for the same set of characters.

173210:
@dukesrg But both of the ways calls mbstowcs, right?

soarqin:
@dukesrg my mistake, it is C11, and only supported by gcc 5.0, devkitarm is based on gcc 4.9 so that it is not supported yet. also swprintf does not support char16_t either, so it is not good to use char16_t here, 4x times memory is an easier choice.

173210:
@soarqin You can switch with an option.

soarqin:
@173210  but gcc 4.9's C11 don't have char16_t in its headers, otherwise we should use c++, that would increase target binary size a lot

173210:
@soarqin My language was insufficient. You can switch the width of wchar_t with an option.

dukesrg:
@173210 Right, but it is called just once on load. Though on a larger set of characters but anyway ANSI is converted more easily.
Switching to 16-bit wchar_t makes a lot of warnings becasue ARM-EABI libraries build with 32-bit wchar_t. That's also points one more time that wchar_t itself is faster on ARM then multibyte just because of nature of ARM CPUs

173210:
@dukesrg So you think ANSI conversion have less overheads than calling the function several times (which should be less than the number of the ANSI characters)?

soarqin:
Anyhow, we are writing an open source project, we use the third-party open source project 'jsmn', I think it is really not a good idea to change it a lot just for unicode support with slight performance or memory improvement. We can solve it by other ways, even they costs a bit more, coz performance and memory are both not the bottleheck of rxTools.

dukesrg:
@173210 Conversion from UTF-8 characters of ANSI is just zero-extending 8-bit value to 32-bit. Calling function even inline will take several more register writes

soarqin:
@173210 ahhhh yes, I forgot '-fshort-wchar' so I misunderstood you

dukesrg:
@soarqin 
https://github.com/roxas75/rxTools/blob/master/rxtools/source/lib/jsmn.c#L8

173210:
@dukesrg swprintf calls mbtowc internally, so it should be slower.

soarqin:
@dukesrg I don't mean code changes break any license, but we are easy to follow origin project's change if we don't change it a lot.

soarqin:
@dukesrg if you have attended a lot of open source projects, you should understand why I consist not to change 3rd-party sources, it is really an annoying thing for open source development.

173210:
So, ok. @dukesrg, you are trying to change many codes to remove such small, ignorable overheads and one man (me) claims it just increases overheads.

dukesrg:
@173210 Not trying but already did.
But anyway. Let's just include this fork
https://github.com/chrisdew/jsmn/commits/master
as a subproject and problem will be solved :)

173210:
@dukesrg What is the "problem"?

dukesrg:
@173210 problem to decide which is better. I'm OK to fallback to original JSMN souce if we jus include it as a subproject.

soarqin:
I don't think there are overheads, thinking just few counts of cpu cycles in modern CPU is totally of no sense

173210:
Ok, so now I have my own working code. I'll commit this.

soarqin:
https://github.com/chrisdew/jsmn
is much older than original project, I prefer port it from bitbucket to some one's own account in github

dukesrg:
@173210 you should mention that severla hours ago :) This PR actually means nothing more that I was just in progress

Syphurith:
@dukesrg @173210 I have some questions for you both, about the JSON, and Unicode.

* Does parsing differs much, with char `"` and wchar_t `"`? If behavior is almost the same, but it is not too easy to judge that is `"` or not, you may even use UTF16.
* Do you really care about the translated text? Those would be simply translated to display. Yes you may want to know the variable name, and these are common chars (i mean 0-127) isn't it? Once you could deal with it similar as above this is almost solved.

Hope this gives you some ideas about it.  Or see **[Here](https://github.com/miloyip/nativejson-benchmark/blob/master/README.md)** for a long list of JSON parsers in C/C++.

dukesrg:
@Syphurith I just need wchar_t* string to draw unicode characters.

173210:
See #110

dukesrg:
@173210 mbtowc?

173210:
@dukesrg yup.

dukesrg:
@173210 I mean with chinese characters?
UTF-8 is the default one so setlocale is omitted?

173210:
@dukesrg right.

dukesrg:
@173210 but.. you know the substring length, why converting each character?

173210:
@dukesrg We have 3 options.
* Add NUL character at the end
A very dirty hack.
* Use `swprintf(strings[j], STR_MAX_LEN, "%*s", s, len);`
Somehow it doesn't work.
* Do it ourselves
It works fine and it's not a dirty hack.

dukesrg:
A very dirty hack that works faster, yes.

173210:
@dukesrg But it's not true in this case.

dukesrg:
@173210 
len = t[i].end - t[i].start; 
for (s = buf + t[i].start; len > 0; s += k) { 
	k = mbtowc(w, s, len); 
	if (k < 0) 
		break; 
 
	w++; 
	len -= k; 
} 

is faster then just

w[mbstowcs(w,buf + t[i].start,len)]=0;

?

173210:
@dukesrg I fixed a bug in the patch. You may understand why I do so if you see f248ad0c02c7ac530996e4a14685033f11330ade.

dukesrg:
@173210 well NM if it is working now, I'm going to change the translation content a lot tonight :)

173210:
@dukesrg I just removed all translations from rxTools.dat because it allows us to implement with a simple code. It is not so good to add more features only for the case.

dukesrg:
@173210 I did the same! But we need the default text with is used if unicode font and/or translation files are missed. ASCII font is integreted now, I suggest we need to make one default laguage JSON file with ASCII-only characters and put it in the rxtools.dat with FilePack..

173210:
@dukesrg If the default language file (en.json) is missing or corrupted, the installation is wrong. You can add a message like "Your installation is corrupted" if you want. I have not added because I'm too lazy.

Syphurith:
@dukesrg Eh.. you're right. I was thinking UTF8 encoded file could be directly converted to wide string (memory), that's what i have mistaken.. (An ascii encoded default language en-GB? LGTM)

dukesrg:
@173210 That is because I want to itegrate default language so rx can work anyway with no language and unicode font.
@Syphurith UTF-8 ecoded file can be converted after loading by mbstowcs, like I was trying to make in this PR, however it requires json parser porting to wide string support too.

173210:
@dukesrg Yes, I know what you want, but actually it doesn't make difference. It just changes the number of files and increases code. It often sounds good that a program is standalone and it does many things, but it's not true.

dukesrg:
@173210 So we can also remove ASCII-part font from the header :)

173210:
@dukesrg But it's necessary to show an error message for font loading (and language file loading) .

dukesrg:
@173210 so we need to keep several hardcoded strings anyway

173210:
@dukesrg Yes, like this.
https://github.com/roxas75/rxTools/blob/master/rxtools/source/main.c#L44
```C
		DrawString(BOT_SCREEN, L"Font load error", FONT_WIDTH, SCREEN_HEIGHT-FONT_HEIGHT, RED, BLACK);
```

dukesrg:
Obsolete now. Will open another PR for fixed translations

Issue #107, opened 2015-08-02
FTP commands isn't work with NetKit ftp client


Nyoroon:
FTP commands in [rxtools/Makefile](https://github.com/roxas75/rxTools/blob/master/rxtools/Makefile) isn't work with [NetKit](http://www.hcs.harvard.edu/~dholland/computers/netkit.html) ftp client:
```
$ make
Name (mirrorcatalogs.com:nyoroon): Please login with USER and PASS.
SSL not available
Password:
convert: unable to open image `unifont-8.0.01.bmp': No such file or directory @ error/blob.c/OpenBlob/2701.
convert: no images defined `gray:font.tmp' @ error/convert.c/ConvertImageCommand/3210.
Makefile:124: recipe for target 'font.bin' failed
make: *** [font.bin] Error 1
```

Nyoroon:
So why not wget?

dukesrg:
@Nyoroon there is no get in devkitPRO/msys package. I'd gladly use curl/wget but we have a lot of windows devs which will be much more unhappy about not working font download than more experienced linux -based. But you're right, as soon as I added different download method for Windows/devkitPRO, we can use wget here for normal download!

soarqin:
We can write a python script to fetch it

dukesrg:
@soarqin I'll fix it with wget for linux builds.

173210:
@dukesrg We can just include the file.

Issue #85, opened 2015-07-20
Improvement: Unicode font support


dukesrg:
It will be great to add a Unicode raster proportional font.
This will also allow the theme system redesign, as text will be a common resource with several languages support. Height of 12px will more readable and easier to optimize drawing with 3DS X/Y swapped screen.

dukesrg:
Memo: http://sourceforge.net/projects/fontforgebuilds this tool can be used for font editing/converting in case we found a suitable font

5685C4A059D5:
Seconding. Seems very useful.

dukesrg:
Used old version with DSTwo font, while can't open it now :(
Need to find free/opensource font with height=12px

Syphurith:
@dukesrg There is `wqy` ([WenQuanYi](Wenq.org)) that known as the default font sets opensource for many linux distributions. If you need to generate the bitmaps, I suggest you to choose a Monospace one.

dukesrg:
@Syphurith the goal is to use proportional as it looks more natural. With 3DS flipped X/Y framebuffer coordinates it is not a big deal to use proportional

Syphurith:
@dukesrg Well. There are normal fonts it offers too. Or, there are opensourced fonts, on github.
Do you want to use TTF with the library, or just a bitmap one that might be faster?

dukesrg:
Bitmap will be easier and faster

Syphurith:
@dukesrg proportional? algorithem? I know there are some embeded solutions that uses bitmaps for different sizes, simply. Even 3DS itself may be just using different bitmaps without calculation with sizes.

dukesrg:
Yes, proportional, in spite of monospace, means each symbol may have a different width in pexels, in case of bitmap, or in units in case of a vector on. Not much of use with Chinese characters but will be more readable with Latin

Syphurith:
Oh you would meet small problems regarding Chinese or even just Japanese characters, or Korean.
Or you would like to seperate the drawing operations for CJK and Latin?

dukesrg:
Which problem?

Syphurith:
@dukesrg when you tried to get a smaller font bitmap from a bigger one..

dukesrg:
@Syphurith you mean scaling the existing font to necessary size? I got what you meant, we will have a problem downscaling chineese/korean/japanese codepages, so we need 8 or better 12 pixel font

dukesrg:
[...]

PR #110, opened 2015-08-03
Fix issues + Improvements

173210:
Solves issue #85 and #107.


dukesrg:
@173210 Well, I've got clean strings with fresh clone :(

rxtools/source/lib/lang.c
```diff
@@ -0,0 +1,96 @@
...
+		return r;
+
+	for (i = 1; i < r; i++) {
+		for (j = 0; j < STR_NUM; j++) {
+			s = buf + t[i].start;
+
+			len = t[i].end - t[i].start;
+			if (!memcmp(s, keys[j], len)) {
+				i++;
+				len = t[i].end - t[i].start;
```
dukesrg: Something is wrong here, though I can't notice it. mbstowcs in place of a loop works fine for ASCII and gives wrong length for UTF, as expected. But the existing source not giving any string.

Syphurith: @dukesrg I don't know what troubles you, but i suggest you to implement a testcase for yourself with desktop GCC. Thus you would be easier testing some implementation, cause you could use netbeans/eclipse/other help you debugging that. At least there isn't any 3ds related code here..

nastys:
I was thinking that we could remove the language selector in Settings and load the JSON file from the theme's folder (after all, the language of the theme should match). If some strings are missing, then it would use the hard-coded, English strings instead.
@dukesrg What do you think?

dukesrg:
@nastys only if theme engine is rewritten completely allowing to specify text/image placing. Otherwise common translation files will be used in all themes anyway.

173210:
@dukesrg It works fine on my 3DS. Your old system.json may cause the problem. Remove it and try again.

By the way, in my opinion, theme dependencies on language should be removed.

173210:
@dukesrg The issue on your 3DS may be fixed with commit 53eacff33d11f795ca351ad9b4fad9a336d8dc01.

rxtools/Makefile
```diff
@@ -125,16 +125,6 @@ font.bin: unifont-8.0.01.bmp
 	@../tools/font_tool font.tmp $@
 	@rm font.tmp
 
-unifont-8.0.01.bmp:
```
dukesrg: You just ripped the font download forcing user to make it by himself?

173210: Yes, that's right.

dukesrg:
I suppose downloading font don't violate GPL. Well have tons of windows hamsters with no font error issue :)

173210:
@dukesrg Windows hamsters? Kill right now :P
```Makefile
	@(test -s $(DEVKITPRO)/msys/bin/ftp.exe && (cat ftp.txt | ftp.exe)) || (cat ftp.txt | ftp)
```

dukesrg:
@173210 at least half of Pasta seems to be on windows. I think there are only few old farts like me, indifferent to OS they are using. As I said before, I intentionally stayed on windows toolchain, because tgere are few devs on linux, which are able to fix compatibility issues, but much more on Windows with no skills. IMO better to have pathch blanket in collaborated code rather then clean release Soon(TM).
About a font - I was going to fix Linux download with wget on next PR.

173210:
@dukesrg Does it matter to include the picture in this repository? We have many external dependencies and this repository have all of them except git submodule. I think the file should be included as other files are.

dukesrg:
@173210 Yes we can keep inside the repo. I just was worried about repo size, but main payload maker is FBI injections anyway.

173210:
@dukesrg Don't worry. Git was originally developed to host Linux kernel repository, which is far larger than this repository.

dukesrg:
@173210 I'm worrying about git clone time :)

173210:
@dukesrg It's not so large. Make Makefile simple.

nastys:
Why would it violate the GPL to add the bmp file directly into the repo?
And by the way, distributing rxTools with the font in it would still violate the GPL.

173210:
@nastys It won't. there are several reasons.
* rxTools is licensed under GPLv2.

GNU Unifont is also licensed with the same version (+ "any later version" option)

* rxTools just includes the binary.

It forces to license under GPL if it changes the form during linking, but rxTools won't (It `convert`s, but I think such a change doesn't matter.)

* The license of GNU Unifont has an exception.

> These font files are licensed under the GNU General Public License, either Version 2 or (at your option) a later version, with the exception that embedding the font in a document does not in itself constitute a violation of the GNU GPL.

nastys:
So, we're not including the font only because it's large. I'll convert it to PNG and use the Makefile to reconvert it to BMP at the right time using ImageMagick (which is needed anyway to convert the theme).

173210:
@nastys It doesn't matter. As I explained.

nastys:
@173210 Then, what about adding it to the repo in BMP?

173210:
@nastys It's completely OK in the same way.

dukesrg:
@nastys if you convert it to png you don't nee dto convert it back to BMP at build time, because I'm anyway converting it from BMP to plain binary with ImageMagic, and only then rearranging glyph data with font_tool.

173210:
By the way, why don't you merge? :)

dukesrg:
@nastys if you're confused with dublicate PR #105 it's ok, result it very similar for enduser. Anyway I've already merged from @173210 fork and working on translation integration. So this PR may be merged, just check chinese/russian are displayed correctly, and that language menu don't allow to click outside the last one (I could push -> several rimes after chinese and then need to push <- to return back to menu)
But just hold with any further translation files or strings changes inside the rxtools directory, it may take several days to fix all the literals.

rxtools/source/lib/lang.c
```diff
@@ -0,0 +1,95 @@
...
+	r = jsmn_parse(&p, buf, len, t, tokenNum);
+	if (r < 0)
+		return r;
+
+	for (i = 1; i < r; i++) {
+		for (j = 0; j < STR_NUM; j++) {
+			s = buf + t[i].start;
+
+			len = t[i].end - t[i].start;
+			if (!memcmp(s, keys[j], len)) {
```
dukesrg: @173210 
This one will false trigger if *s is a substring of keys[j] since length of keys[j] is not considered.

173210: If `len` is longer, it won't match with a NUL character.

dukesrg: Dunno, just failed with having both
BUTTON_A and BUTTON_ACTION
this is the first suspect I found.

173210: @dukearg Actually it fails. My JSON readers are the minimalist implantation.
If you want to have them, add `keys[j][len] == 0`

dukesrg: @173210 where exactly?

173210: @dukesrg After (not before) `!memcmp(s, keys[j], len)`

dukesrg: @173210 why to trim the existing key?

173210: Not trimming. I wrote `==`.

dukesrg: @173210 my bad :(
But it will be better like that
!keys[j][len] && !memcmp(s, keys[j], len)

173210: !key[j][len] should be after memcmp, although it's not so different.

dukesrg: If key[j][len] is not null, we should not compare the rest because s and key length are different and we need them to be exact equal to proceed with value copy.

173210: We don't know `len`, so we'll check `len` < The length of the key by `memcmp`. It prevents illegal access (Anyway, I don't think it matters.) .

PR #117, opened 2015-08-05
Remove tools

173210:


dukesrg:
Wow, finally it all moved to GNU tools :)
BTW enven patches is .elf now, the content is still just a stripped code, right?

173210:
@dukesrg Yes, but the file is a little bigger than roxas75's format because ELF has many variables. But the format for general purpose may be better than roxas75's format for the further extension.

PR #120, opened 2015-08-05
fix reboot.bin and some clean optimizations

Matsumot0:


Makefile
```diff
@@ -72,7 +73,7 @@ all-target-brahma:
 	$(MAKE) -C brahma
 
 reboot/reboot.bin:
-	$(MAKE) -C $(dir $@) $(notdir $@)
```
173210: Can you explain this change>

Matsumot0:
i don't remove this entire line, but just $(notdir $@) variable to avoid  compilation error on reboot.bin file due of nastys change nothing else.

173210:
I looked at reboot/Makefile. The crappy devkitPro format seems to prevent the recursive make. Anyway, the change makes sense now.

Matsumot0:
:+1:

PR #122, opened 2015-08-06
Update rxtools/Makefile and transpose the font

173210:


dukesrg:
@173210 better now :+1:
I'll check once again if it possible to easy rearrange font image with imagemagick to simplify glyph address calculation.

173210:
@dukesrg Actually I know a better command. But I'm looking for the better solution...
```
convert unifont-8.0.01.bmp -crop 4096x4096+32+64 -depth 1 -negate -transpose bmp:- | convert -crop 16x4096 -append bmp:- gray:font.bin
```

AlbertoSONIC:
Can I merge?

173210:
@AlbertoSONIC Yes, of course.

dukesrg:
@173210 if -append writes to the end for gray format, why not? It will make a linear incrementing font data. Most probably you won't avoid using pipe for that operation anyway, so looks fine

173210:
@dukesrg By the way, does pipe work on Windows? Here is my latest command.
```
convert $< -crop 4096x4096+32+64 -depth 1 -negate -transpose - | convert -crop 16x4096 -append - gray:$@
```

173210:
So whether this works or not.
https://github.com/173210/rxTools/commit/69609178658dd0dc0234b40c944635e19c98b078

dukesrg:
@173210 it works fine from the shell, but make breaks the pipe

dukesrg:
@173210 Oh, wait, its working just fine!

173210:
@dukesrg Review #123,

dukesrg:
@173210 Can't find comments from that PR anymore, so I'll repeat here also just in case. This one will change byte order to LSB first, so whole glypth column may be shifted down as one unsigned short with screenbuffer address(y coord) increase
@convert -crop 4096x4096+32+64 -transpose $< - | convert -crop 16x4096 -append - - | convert -roll +8 -negate -depth 1 - gray:$@

PR #129, opened 2015-08-06
System dependencies

173210:
I moved rxTools.dat to rxTools/system/code.bin because now rxTools/system has system dependencies.
Users can update rxTools by copying them to rxTools/system directory.

Entry points need updated, but I have only tested spider entry point.


AlbertoSONIC:
Are you sure it is a good idea to move everythig to rxtools/system? I mean, there are many noob users which may not like to have to deal with many bins

173210:
@AlbertoSONIC I know that, but actually we have several files. Then we may have better have only one directory.
And even users think many bins are not good, we don't have any disadvantage except the fact that they think so. :)

AlbertoSONIC:
@motezazer @capito27 @nastys @dukesrg @ApacheThunder @LuxSyphurithSIe I'd like to know your opinion

AlbertoSONIC:
@173210 I mean.. there are many users who prefer to use Reinand instead of rxtools since reinand doesn't require the firmware.bin! (it's inside the .dat)

173210:
@AlbertoSONIC Well, but actually it's troublesome to download firmware.bin, but it doesn't matter in this case.

AlbertoSONIC:
@173210 It's better to have 1 directory only, instead of a directory + rxtools.dat.... but it's even better to have an rxtools,dat only!

ApacheThunder:
Hmm...I'm not sure about this. But for me it would be easy to adjust to assuming the MSET roploader is updated accordingly. It should be hard to update my browser URL. The same can't be said for the inexperienced users though. It might throw them for a loop for awhile if you decide to move rxTools.dat to to another folder. I assume you also intend for the firmware.bin to be located elsewhere as well?

173210:
@AlbertoSONIC It doesn't matter because we have already several files.

AlbertoSONIC:
@173210 This is definitely true. On pasta we had an awesome Companion App, which downloaded and decrypted the firm with a click of a button. Maybe we should port that companion app to rxtools

173210:
@ApacheThunder I suggested this change because entry points need updated for a decrypted version by CakeHax.

173210:
@AlbertoSONIC Yes, that's a true solution.

AlbertoSONIC:
Maybe we could create a Companion App which prepares the sdcard with everything needed. The user selectes the drive, then clicks INSTALL. It downloads the rxTools folder and puts it to the sdcard, then it downloads the firmware.bin and it puts it into the sdcard. Done and noob proof!

173210:
We may have an enhancement issue for that.

AlbertoSONIC:
In the meanwhile, i merge this?

173210:
Yes, in my opinion. ;)

nastys:
@AlbertoSONIC Didn't Nintendo remove everything from their servers? How is the companion app supposed to download the firmware? :sweat_smile:

AlbertoSONIC:
@nastys Someone could host a mirror ;)

173210:
@AlbertoSONIC It's piracy. Supporting the latest firmware is the only legal way.

dukesrg:
Well, actually it is possible to download files from spiderhax so virtually SD card management won't be used at all having spiderhax available. And increasing the number of required files will make it just harder to implement.

173210:
@dukesrg It's not increasing except font.bin. ;)

dukesrg:
@173210 I meant, having to download only rxTools.dat is mucheasier than to download whole /rxRools with subfolders. Expecially with no actual ability to create folders under spiderhax.

173210:
@dukesrg I got it, so you're talking about an implementation to download not only firmware, but rxTools.

dukesrg:
@173210 yes, why not!

173210:
@dukesrg Well, we need a downloader, but, yes, actually spiderhax can be.

AlbertoSONIC:
I prefer the Companion App way. The downloader can work with Spider and Ninjhax, but what about MSET? The companion app i'm working on has rxTools install/update feature (incl firmware), mset downloader and fbi injection related file downloader. And the only way is using a mirror. I know it's piracy, but is there any other option?

173210:
@AlbertoSONIC Nintendo should continue to host the latest firmware. We don't have to use any mirror.

AlbertoSONIC:
What about MSET?

173210:
@AlbertoSONIC Oh, so you are talking about MSET downgrading?

AlbertoSONIC:
Yes

173210:
@AlbertoSONIC Well, MSET downgrader is not alyways necessary, so it can be ignored, I think. And, in my arrogant opinion, it's not better than ninjhax.

173210:
By the way, why don't you open an new issue?

AlbertoSONIC:
For the companion? I'm an n3ds user so I'm forced to use ninjhax. But many people use mset and put an mset downloader into the companion would be nice...

173210:
@AlbertoSONIC But is it worth piracy?

AlbertoSONIC:
Not really.... So there's no solution.. No 9.5 firm, no mset. Nintendooooooo :(

nastys:
@AlbertoSONIC If having multiple files is the issue, then we could put rxTools.dat INSIDE the rxTools folder and make rxTools check inside it for firmware.bin and slot0x25keyx.bin if they cannot be found in the root.

If you manage to boot the latest emuNAND on N3DS, we'll not need anything Nintendo removed.

As for the O3DS, I made a generator that is very similar to MakeKeyBin (Wii), but somehow these kind of generators are not allowed on GBAtemp.net... (you can still find it on Reboot.ms)
O3DS doesn't need firmware.bin.

dukesrg:
@nastys how about TWL/AGB firm? Aren't they necessary for those modes to patch?
I was thinking placing such files somewhere on the SD card subfolder like /cdn/ or /nus/ and recreate cdn folder structure to make those files shared with different exploits. It will allow to have one universal downloader companion app, and meybe spiderhax downloader (well there is till create directory not available for now).
As for having the minimal needed data to boot inside rxtools.dat (with filepach or anything), that may be an option. We already have ASCII font and I also kept several urgent strings statically initialized in translation so that fork should boot fine with no translation and unicode font. So if we're to keep some data inside, at least ascii font and simple translations can also be moved into filepack (or whatever) as a binary resources.

soarqin:
mset exploit support is broken after this PR is merged, I think mset exploit does not support to boot files inside any subfolders

173210:
@soarqin Have you updated your entry point? If so, I think it's just a problem of the length of path.

soarqin:
@173210 of coz I did, I reinstalled by new generated rxinstaller.nds and then gave another try with ropCustom.txt, both failed

173210:
@soarqin I found rxInstaller is corrupted. I'll fix it...

soarqin:
@173210 okay, so CakesROPSpider is also not working?

173210:
Yes, because rxinstaller doesn't pass the correct path for it.

173210:
@soarqin See #138.

PR #134, opened 2015-08-08
fixed CTR decryption (once again)

dukesrg:
That one probably will be the last fix for decryption :+1:


ApacheThunder:
Yep. I tested this commit from his fork and succesfully decrypted romfs of home menu cxi and tested a large game like Pokemon Alpha Sapphire. Unlike before, I no longer get corruption of the large romfs.bin. It passes all 3 level checks using the ctrtool -y command. Safe to say CTR Decryptor is free of major issues now. :D

PR #138, opened 2015-08-09
Fix commit d9f9e41b78326a078dcf8bdb0cc6330205c5ec30

173210:
MSET works now on my 3DS.


soarqin:
good job, would try it later

173210:
@soarqin This is my fault. Sorry for the problem.

PR #153, opened 2015-08-10
rxmode/native_firm

173210:


nastys:
@AlbertoSONIC Why not Sysn/Emun, or Ver+/RX-E, or even RSYS/REMU?

AlbertoSONIC:
RX3D and Ver+?

173210:
So that can distinguish rxMode with normal mode?

AlbertoSONIC:
Yes

173210:
In my humble opinion, the current behavior that rxTools boot sysNAND when emuNAND fails may be a far more serious problem.
Users may say the display is bad, but they may have tried to boot emuNAND which is not installed and jumped to sysNAND :P

AlbertoSONIC:
It boots sysnamd but also sets the label to "Ver." which is enough for the user to know that the emunand hasn't been booted.

173210:
But if I were a noob, I would want to tell me when it fails to boot emuNAND.

AlbertoSONIC:
ConsoleInit();
ConsoleSetTitle("EMUNAND NOT FOUND");
print(L"The emunand was not found on \n");
print(L"the sdcard. \n");
print(L"Press A to boot the sysnand \n");
ConsoleShow();

173210:
@dukesrg It's not a good idea to do several things in a commit. And don't worry, I just pushed 25 commits a day.
But, actually we may have too many merging commits. Using git --rebase or giving me a writing permission is the easy solutions.

PR #155, opened 2015-08-10
more smooth fade-out

dukesrg:
@AlbertoSONIC no your decision - a bit smoother fade-out effect.
may be even smoother/faster if load image into separate buffer and read source from it instead of framebuffer.


AlbertoSONIC:
Why loading the image on another buffer?

Syphurith:
@AlbertoSONIC Double buffer would decrease blinking. Also load to VRAM slower than just to RAM.

dukesrg:
FYI that weird shift-and-add stuff is a 32-bit packed 8 bit multiply by 0.96875 which took 64 iterations to make less then 1 from 255

PR #165, opened 2015-08-12
screenshot improvements and some minor fixes

soarqin:
Old screenshot function captures bad image due to some typo and GetPixel are somewhat incorrect.
I didn't look into the GetPixel issue but just stripped use of GetPixel and read VRAM directly.

### DIFF
--- a/brahma/data/payload/source/main.c
+++ b/brahma/data/payload/source/main.c
@@ -4,7 +4,7 @@
 
 void* (*rxTools)() = (void*)0x08000030;
 
-void main()
+int main()
 {
 	/*int fbfound = 0;
 	unsigned char* screen = 0x20000000;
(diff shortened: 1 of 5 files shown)


173210:
:+1: But you shouldn't use File* and f_* at the same time.

# In my opinion, File* should be removed. They just ignore the return values.

soarqin:
I used FileOpen just because I was too lazy to do some checks that already written in f_open.
I am going to add a commit to remove them.

173210:
@soarqin In this case, You have already used f_*, so you shouldn't use them.
By the way, you used `f_sync`, but it doesn't make sense because it calls f_close just after that. f_close automatically flushes the buffer and saves correctly.

/* This is another reason why File* should be removed. f_sync is usually unnecessary. */

soarqin:
@173210 f_flush is just called in my intermediate codes which used to keep a loop cache to write the bmp file multi times, after that I used a complete block of memory to hold the whole bmp body so that it is no more needed, I forgot to remove it. Thanks for tipping

PR #166, opened 2015-08-12
Revert "Fixed Ver. string patch"

173210:
FCRAM_ADDR_MAX is not the true bottom of FCRAM.


AlbertoSONIC:
But it was working :(

173210:
@AlbertoSONIC It should be the bottom of MSET (or very close, at least) . The address is only valid for the patch.

AlbertoSONIC:
:+1:

PR #189, opened 2015-08-14
Move firmware.bin to rxtools/system

ApacheThunder:
I tested this, It works. Might as well have it read the file here since just about everything else is being loaded from this directory now. :D

### DIFF
--- a/rxtools/source/features/configuration.c
+++ b/rxtools/source/features/configuration.c
@@ -260,7 +260,7 @@ int InstallData(char* drive){
 	f_mkdir(path);
 
 	//Read firmware data
-	if (f_open(&firmfile, "firmware.bin", FA_READ | FA_OPEN_EXISTING) != FR_OK) return CONF_NOFIRMBIN;
+	if (f_open(&firmfile, "rxtools/system/firmware.bin", FA_READ | FA_OPEN_EXISTING) != FR_OK) return CONF_NOFIRMBIN;
 	wcsncpy(progress, strings[STR_PROGRESS_OK], wcslen(strings[STR_PROGRESS_OK]));
 	progress += wcslen(strings[STR_PROGRESS_OK]);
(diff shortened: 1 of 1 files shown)


Syphurith:
If the file is only used here, then a new path does support you this. LGTM.

ApacheThunder:
Hmm the broken English of that sentance is making it hard for me to read what you are trying to say. Can you reword it?

I removed data directory and recompiled with the above change and moved firmware.bin to that location. It seems to work fine for me. It's really just a suggested change to tidy up root directory of root of SD card as that was the direction some of the past commits were doing. :P

Syphurith:
@ApacheThunder Yup. My english is not good at all. LGTM - Look good to me.
You may try to search whole project for string "firmware.bin" if there isn't any other, yes that's good.

ApacheThunder:
I see. :P

I just now noticed this causes the installation screen to come up on every boot. GBA VC/DSI stuff still works, so it does create the bin files correctly. But perhaps there's an additional check somewhere it's doing and it thinks it needs to do the install again even though it doesn't. Unless this is a issue with the current build already? I don't see how moving the firmware.bin file could cause this, but I'll take a look to see if I missed anything.

ApacheThunder:
Ok even without my change, it still fails to create the bin file for agb_firm (I'm referring to the bin files it creates for each firm in the rxtools/data folder). This seems to be a pre-existing issue. My change didn't cause this as it was already occurring.

Syphurith:
@ApacheThunder If it could create the patched AGB firm before.. i don't know then.
I grepped it and find just two appearence of "firmware.bin", in cdn_firm.py, and configuration.c.
*..I am hoping for a SEEDDB dumper feature.. Too lazy i am..*

ApacheThunder:
the py script creates the bin file at the location the python script is located. The end user was always expected to move the bin file to the correct location on the SD card themselves after using the python script.

PR #192, opened 2015-08-14
FBI Inject translation updated

dukesrg:


lang/it.json
```diff
@@ -8,7 +8,7 @@
 	"DUMP": "Estrai %ls",
 	"DUMPING": "Sto estraendo %ls to\n%s...\n",
```
nastys: If you copied the "to" part from here, I forgot to translate it :grin:

dukesrg: Noticed that in both it and es files :)

nastys: That's because I translated in both languages :stuck_out_tongue_winking_eye: 
It's actually "en" in Spanish and "su" in Italian.
Anyway, I've already fixed that in the latest commit :wink:

PR #202, opened 2015-08-15
fixed firm patches

### LINKED ISSUE #190: AGB Mode appears to be broken...
Current build appears to have broken AGB_FIRM. It won't boot and data install screen comes up everytime. Did you change structure of firmware.bin again? If not something broke. It's never creating the agb_firm bin file in the data folder. This breaks AGB mode as well as causing the data install screen to come up everytime rxTools is booted. It does create TWL and Native_Firm bin files correctly, so TWL mode and rxMode still work.

soarqin:
This should fix #190

### DIFF
--- a/rxmode/agb_firm/script.ld
+++ b/rxmode/agb_firm/script.ld
@@ -1,5 +1,9 @@
 SECTIONS
 {
+	.text : { *(.text) *(.text*) }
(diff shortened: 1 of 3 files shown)


soarqin:
I don't know why SECTIONS in ld script requires default sections (.text/.bss/.data) prepended, but at least it works after I added them

Issue #203, opened 2015-08-16
code.bin is broken


kot2002:
It seems that CakesROPSpider can't handle changed file path. (/rxTools.dat -> /rxTools/system/code.bin)

PR #204, opened 2015-08-16
Added CakesROP and update CakeHax

kot2002:
I think it is more reasonable to use CakesROP instead of rxinstaller.
This solves Issue #203. Sadly, I have not tested this on new 3ds.


AlbertoSONIC:
Yeah someone reverted my "Update CakeHax" commit... Thanks ;)

PR #205, opened 2015-08-16
Remove rxinstaller and relocate logo

kot2002:


soarqin:
good to see wo do not need to port code changes from CakesROP any more, but we may need to make a PR to CakesROP to fix warnings in CakesROPSpider which was done by @dukesrg

PR #210, opened 2015-08-18
Revert bigpayload offset

kot2002:
mid-kid/CakeHax@284052f183cef8a548ffeae4baf5348129a1c7c8

This changed the offset.


kot2002:
Because the name of rxTools.dat was changed to code.bin recently, 
I changed the name of another code.bin (spider mset installer) to rxinstaller.bin to avoid confusions.

PR #220, opened 2015-08-24
Fixed region-free patch

soarqin:
9.9.0TWN should work now.
I don't know if 9.9.0USA was broken or not, but this should fix 9.9.0USA if it was not working.

### DIFF
--- a/rxmode/native_firm/source/payload/arm9/myThread/myThread.c
+++ b/rxmode/native_firm/source/payload/arm9/myThread/myThread.c
@@ -209,7 +209,7 @@ static void findRegion()
 {
 	uintptr_t p;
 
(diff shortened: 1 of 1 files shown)


Syphurith:
@soarqin I have digged into NTR Language Emulation plugin. The plugin solves out-of-region DLC issue. If you have some abilities and would want to take a look, i could give you the url.

soarqin:
@Syphurith The resolution of language emulation plugin has a critical problem that we are hard to identify the apps' region, only if we write a config file to specifiy them. I think we can look out NTR's home menu region-free patch, which is different of rxTools, but it does not emit black blocks on CHN/TWN consoles

PR #228, opened 2015-09-01
Fixed MSET DG

kot2002:


rxtools/source/features/downgradeapp.c
```diff
@@ -271,7 +276,7 @@ int checkDgFile(char* path, unsigned int hash)
 	{
 		rb = FileRead(&fp, buf, fixedsize, 0);
 		FileClose(&fp);
-		if (HashGen(buf, rb) == hash) return 0;
```
Syphurith: Orz. I was wondering the expressions are almost the same.. And turns out `0` means failed here.. So `if (HashGen(buf, rb)!=hash) return 0;` should do the same.

kot2002:
Pulled new request!

PR #233, opened 2015-09-03
Change BOT_SCREEN2

kot2002:
It fixes N3DS MSET.
Tested with O3DS 9.0E SPIDER/MSET(6.x DG)/NINJHAX, 4.5E SPIDER/MSET, 4.5K SPIDER/MSET.
Entries I tested were not compromised by this commit.


nastys:
Fixed properly ;)

PR #235, opened 2015-09-05
Support SDinfo.dat and ncchinfo.dat from 3DS_Multi_Decryptor

soarqin:
3DS_Multi_Decrytor uses unicode filename in SDinfo.dat and ncchinfo.dat, I did a liitle trick to detect them.

### DIFF
--- a/rxtools/source/features/padgen.c
+++ b/rxtools/source/features/padgen.c
@@ -155,7 +155,18 @@ uint32_t CreatePad(PadInfo *info, int index)
 	#define BUFFER_ADDR ((volatile uint8_t*)0x21000000)
 	#define BLOCK_SIZE  (4*1024*1024)
 
-	if (!FileOpen(&pf, info->filename, 1))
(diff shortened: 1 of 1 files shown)


rxtools/source/features/padgen.c
```diff
@@ -155,7 +155,18 @@ uint32_t CreatePad(PadInfo *info, int index)
 	#define BUFFER_ADDR ((volatile uint8_t*)0x21000000)
 	#define BLOCK_SIZE  (4*1024*1024)
 
-	if (!FileOpen(&pf, info->filename, 1))
+	if(info->filename[0] != 0 && info->filename[1] == 0 && info->filename[2] != 0 && info->filename[3] == 0)
+	{
+		char fname[90];
```
dukesrg: Should be sizeof(...) instead of 90 in both cases

soarqin: yup, I changed the codes in #236

PR #244, opened 2015-09-13
Removed hard-coded FirmInfo values and generate them from FIRMs

soarqin:
I have tested it on my O3DS.
But I do not have any N3DS, anybody who holds N3DS please check if this works for N3DS too.

This is the 1st step leading to support different versions of NATIVE_FIRMs, but we stiil need to find a way to change patch offsets on the fly.


173210:
#258 has been merged, so this pull request should be closed.

PR #247, opened 2015-09-26
Ported rxTools to brahma2

WilliamO7:
So in light of the port of brahma to *hax 2.0 based exploits by delebile (https://gbatemp.net/threads/porting-brahma-to-ninjhax2.398243/), I took the liberty of porting rxTools to this new code, and lo and behold, it boots successfully on my 9.2.0-20 EUR 3DS.

With this, we can now (with skillful payload running) autoboot rxTools using homemenuhax (and by proxy, other CFW solutions.)

This will need some tidying up though. And it only has an 80% boot rate so far. And it only works on O3DS. (Yes, it is quite obvious I'm jumping the gun here.)

### DIFF
--- a/brahma/source/arm11.s
+++ b/brahma/source/arm11.s
@@ -96,7 +96,7 @@ wait_arm11_loop:
 	BX              R1
 
 	pa_hijack_arm9_dst:  .long 0x1FFFFC00
-	pa_arm11_code:       .long 0x1FFFFFF8
+	pa_arm11_code:       .long 0x1FFFFFFC
 	pa_pxi_regs:         .long 0x10163000
 	some_pxi_cmd:        .long 0x44846
 	pa_firm_header:      .long 0x24000000
--- a/brahma/source/arm11.s
+++ b/brahma/source/arm11.s
@@ -96,7 +96,7 @@ wait_arm11_loop:
 	BX              R1
 
 	pa_hijack_arm9_dst:  .long 0x1FFFFC00
-	pa_arm11_code:       .long 0x1FFFFFFC
+	pa_arm11_code:       .long 0x1FFFFFF8
 	pa_pxi_regs:         .long 0x10163000
 	some_pxi_cmd:        .long 0x44846
 	pa_firm_header:      .long 0x24000000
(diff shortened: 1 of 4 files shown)


ApacheThunder:
Oh nice! This should hopefully get the bugs ironed out by the time I get my 3DS back out of the pawn shop early next year. Yet again, something interesting happens when I don't have access to my 3DS. :P

julian-20:
Works fine! (8
You can autoboot rxtools. Just rename rxtools.3dsx to boot.3dsx and place it in root of the sdcard.

kot2002:
Brahma 2 just updated. Does not work on N3DS either.

WilliamO7:
About it not working on the KTR (N3DS) platform, I was thinking that we could have folders with Brahma 1 and 2, so people can use one distribution or the other.

Also, I'm away from my dev PC right now, so I can't do much at the moment

kot2002:
@WilliamO7 
Brahma2 with ninjhax1.1 on N3DS would work. I think the seperation is not necessary.

WilliamO7:
So as you can see above, I updated Brahma, and it now does a sterling job.

AlbertoSONIC:
But not on n3ds, as far as I know :(

WilliamO7:
Anyone tried building this and booting on ninjhax 1.1 on a N3DS to see if this works?

ApacheThunder:
I would hope one could provide a custom themehax payload so we don't have to use the boot.3dsx file name to auto boot it. Some folks may wan to still use Ninjhax and they'd constantly have to switch around the 3dsx files if they wanted to boot into Ninjhax after using brahma2. :P

WilliamO7:
I'm in the process of doing that :) just need to work my way around this asm...

WilliamO7:
Good news guys!!!
I just ported Steveice's CakeBrah to rxTools and it now works perfectly as a boot.3dsx!
Which means cold boot to emunand can be done in less than 15 seconds!
So I'll close this now. Until mid-kid (or b1l1s if he comes back) merge the changes in, it wont officially be here, but you can try it out [here](https://github.com/WilliamO7/rxTools/tree/bruh).

PR #258, opened 2015-10-18
Remove FirmInfo

173210:
This change is almost same with #244. It depends on #257 and removes FirmInfo completely.
Also this pull request doesn't include changes for rxmode in #244.

### DIFF
--- a/rxtools/source/features/configuration.c
+++ b/rxtools/source/features/configuration.c
@@ -236,8 +236,6 @@ int readCfg()
 }
 
 int InstallData(char* drive){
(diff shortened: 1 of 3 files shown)


soarqin:
Is this tested on n3ds? I found firminfo of n3ds is different from generated values so that I did request tests on n3ds in #244
PS: rxmodes changes in #244 are ported form cakes

173210:
@soarqin Please check #259.
Write what commits do in their logs. I'm not sure what rxmode changes in #244 do.

PR #259, opened 2015-10-18
Add PROCESS9_CODE_ADDR

173210:
This will fix the patch on Process9 for KTR as #244 does.
I have not tested, so please check whether this works.


AlbertoSONIC:
My MicroSD adapter just died... @motezazer @capito27 can you test this?

capito27:
Just tested, it freezes when launching both emunand and sysnand on my N3DS (tried from Ninjhax 1.x).

173210:
@capito27 How about before applying this?

173210:
I close this pull request because the PROCESS9_CODE_ADDR for KTR in this commit is invalid.

Issue #240, opened 2015-09-12
Make patches targeting new released 10.0.0-27 NATIVE_FIRM


soarqin:
Nintendo may remove 9.6.0-24 NATIVE_FIRM from their cdn sooner or later like they did to other old firmware titles, so we had better start to make patches for newest NATIVE_FIRM.
As researches from 3dbrew, there is no change to patched parts, there are just changes in address offsets, I think it is not hard to do it.

nastys:
Has this been fixed yet?

soarqin:
nope, I do not know how to dump 10.0.0+ firm memory for patches analyze

nastys:
OK, I don't know how to do it either, so I'll label this issue as enhancement.

PR #264, opened 2015-10-21
Update configuration

173210:
Tested on old 3DS.
e17b0744b29bdaec13f63f769725799bc9459938: because rxTools/system is now rxTools/sys. The short name makes sense.
54bcce5a6dfea587703476e6e8b91d9eb5211f93 allows to add more files (such as CETK) easily. It also removes hardcoded firmware sizes (maybe necessary for #240). New users can download firmware with browsers (just put downloaded files in rxTools/firm directory) .


173210:
Rebased to ca4afc3513412f9702f0dfcd6076c08751678db3.

PR #289, opened 2015-11-06
Update GUI 'import.db' + fr.json + themetool

FONZD:
Some minor updates


Syphurith:
@FONZD Well next time if you find the git log of your branch is polluted, once you solved that, you can simply update the branch again, with "unknown change" so github won't reject it.

Issue #297, opened 2015-11-08
N3DS doesn't load sysnand, emunand home screen partially loads


urherenow:
Now that I have the data installation suite bit figured out, trying to boot sysnand gives me a black screen. When booting emunand, the home menu never quite finishes loading. Sometimes it just freezes, other times I can move around, but the couple of folders that it does show are empty, and the rest of my icons never load.I can currently see (but not load) a few empty folder, AR card game thingy, face raiders, and pokemon rumble world icon. That's it.

EDIT: This is booting through menuhax 2.0 and ctrbootmanager b08
EDIT2: Tried without using a theme on emunand. Also upgraded to ctrbootmanager 1.0 and tried using the yellows8 link for browserhax for good measure. Same results.
EDIT3: Looks like this issue crept in after 1edb3cd <--(last one that doesn't have this issue)

AlbertoSONIC:
This is happening to me too. Right now I'm in the homescreen of the emunand, icons are partially loaded, background music is gone, the banner on the top screen is gone, the "Launch" button isn't showed. I can move in the menu, I tried to take a screenshot of them bottom screen (using the home menu screenshot feature) and the n3ds completely freezed. I guess there's some process9 fs service open...

AlbertoSONIC:
It's basically like this:

-Emunand boots and arm11 loads the theme and half the icons.
-An arm9 process/thread starts using fs, so arm11 stops loading the icons and the bgm stops playing. 

@173210

urherenow:
Almost sound right, but when using the Majora's mask theme, the music keeps playing as long as I can move around the partially loaded menu. I turned off my theme now though to eliminate is as a cause. At any rate, it never completes loading stuff from the mSD card.

AlbertoSONIC:
@173210 I still get the same freeze on my n3ds, even with #299 :(

urherenow:
You have to try with O3DS, as this pull was only for TWL, not the n3ds issues.

AlbertoSONIC:
@173210 said his PR fixed this!

173210:
Hmm, my new 3DS can launch rxMode, and others' can't. So it should be the problem of alignment.
I found the size of a section is not aligned by 16 bits (2 bytes), so it causes the issue, I think.
But I have no idea why it does matter. I used '<' (less than) to compare the top and bottom address, so it should work...
[...]

Issue #284, opened 2015-11-02
Sggestion: Remove CakeBrah


173210:
The reason is quite a simple: there is CtrBootManager. Even the user guide tells to use CtrBootManager for themehax.
https://docs.google.com/document/d/1vLSLAvFW_4UekzJnKCjz_ce7AWYmjp5Cq2LxfQ9cDr0

urherenow:
But what about Ninjhax users? Doesn't rxtools.3dsx still need Cakebrah for that? This would make menuhax/browserhax a requirement.

urherenow:
Or does the homemenu ropbin take care of that? I'm not quite sure what that's for...

urherenow:
Also, the guide talks about using HBL...

173210:
@urherenow It works on Cubic Ninja and I'm using it now. :)

urherenow:
Without Cakebrah?

173210:
@urherenow Yes.

urherenow:
Cloning your tree right now to see if it will build/boot. Without Cakebrah, how is it gaining kernel access?

173210:
@urherenow Oh, I mean I didn't use CakeBrah in rxTools. Of course CtrBootManager includes it.

urherenow:
Did you use Ninjhax to launch ctrbootmanager directly? That's what's confusing me. When I think of Ninjhax, I think of booting into HBL. Then rxtools would have to have Cakehax. If Ninjhax can directly boot ctrbootmanager, then that will need to be updated in the manual (but would work none the less).

173210:
@urherenow Yes, I did so.

urherenow:
Ok, so building on Windows (your current tree), gets stuck with a red bottom screen when directly booting HBL through menuhax 2.0. When booting ctrbootmanager through menuhax, I can get to the rxtools menu, but whether I auto boot emunand, or manually pick SYSNAND, it halfway loads emunand and freezes (shows only a few icons and the rest never finish loading).

And for more clarification, what about booting rxtools via MSET without Cakehax? That works as well?

173210:
@urherenow Wait, we are talking about CakeBrah, not CakeHax, ok?

kot2002:
Well, the CakeBrah submodule has been reverted and is quite stable on Ninjhax 1.1b, so I think this issue could be closed.
[...]

PR #301, opened 2015-11-10
Removal of CakeBrah, revise README 

kot2002:
It's you collaborators' choice whether accept this or not :P


urherenow:
With this commit, we will NEED ctrbootmanager?

kot2002:
Yes. If you are going to use rxTools with *hax.
And actually, current source does not support ninjhax 2.0-2.5 since CakeBrah was reverted to previous version which supports only ninjhax 1.1b.
See this : #284

urherenow:
Wait... Is hax 2.0/2.5 the reason for #297? If so, why wasn't that mentioned? Does the current commit work with Ninjhax 1.1b on N3DS?

kot2002:
Current rxTools.3dsx works with Ninjhax 1.1b only. So we have little reason to have it in the upcoming stable release. (Ninjhax 1.1b? really?)
And CakeBrah that supports Ninjhax 2.x is buggy.
CtrBootManager/BootCtr are working with Ninjhax 2.x and you can use them not only to autoboot, but to run rxTools from HBL as well.

urherenow:
Don't know about right now, but the commit before today is useless with *hax 2.5 and ctrbootmanager. See #297. @kot2002 your replies always confuse me... I'm talking about N3DS.

kot2002:
Did you try with the latest commit?

kot2002:
And I misread your comment. No this is irrelevant with current N3DS issue.

urherenow:
I'm in Tokyo for tonight. Can't try the latest until tomorrow evening (JST). But doubt I want to until a pull request claims to fix #297. And as for you other comment "Ninjhax 1.1b? Really?". Yes, really. It had what seems at the moment to be the most stable kernel access aside from Gateway. The only thing Pasta or Rxtools ever brought to the table for CFW was the sig patching. How the heck could you dis Ninjhax 1.x?

m45t3r:
I am somewhat against of this (and I am the creator of BootCtr, it should say something). For example, without CakeBrah users wouldn't have a way to boot rxTools from HBL itself, since neither BootCtr nor (AFAIK) CtrBootManager includes a .smdh file. While it is possible to include a .smdh in both projects, I still think it is nice to have a dedicated .3dsx file from rxTools itself.

Why rxTools isn't using the lastest commits from CakeBrah btw? It was because my report about bottom black screens? I thought the responsability of that was in CakeHax, no CakeBrah.

Anyway, I send some commits to CakeBrah to fix somethings (most related to warnings and including a better function to load exploits), you could start to use them. Or the problem is the low boot rate from directly booting .3dsx file with themehax is the issue? If so, this should be trivial to fix by including a small delay when loading the payload.

kot2002:
OK. I will close this.

173210:
@m45t3r The current CakeBrah is unstable because of libkhax.

m45t3r:
@173210 So you basically have poor boot rate with Ninjhax 1.x/2.x support or good boot rate with Ninjhax 1.x support only?

And can't you include a small delay in booting like I have in BootCtr (https://github.com/m45t3r/BootCtr/blob/master/source/loader.c#L3), this seems to improve boot rate by a lot. Or unstable is another thing, like crashes or something like this?

AlbertoSONIC:
RxTools always boot with ctrBoot for me, but I often get a black bottom screen...

173210:
@m45t3r It rarely boots with Ninjhax 1.x and even with Ninjhax 2.x. But the previous version worked fine with Ninjhax 1.x.
I don't know well because I don't want to read the source code (libkhax is written in C++!)

PR #324, opened 2015-12-30
Update Makefile

urherenow:
Fix missing leading "/"s in line 18 and 19. Won't build on Windows without them.


urherenow:
This will not be needed if we revert back to a08cfdd

PR #348, opened 2016-01-09
Update build

mariogamer2:
Now that Cake have to be cloned separately,I updated.


urherenow:
No. This isn't how to do it. rxtools needs to start using the latest by default.

AlbertoSONIC:
Yeah. We have to update submodules

PR #359, opened 2016-01-14
Fix rxtools

### LINKED ISSUE #358: Attempting to inject FBI leads to hang, commit 647b341
As background, I'm using an o3DS downgraded to fw 9.2 in sysNAND, 10.3 in emuNAND.

I was able to use an older (not sure which) version of rxTools to inject FBI. When compiling 647b341 and installing it, trying to read and just check FBI on both sysNAND and emuNAND leads to rxTools to hang. I did a `git bisect` and traced the issue to commit 77c1f93773113d85421402193d338fede5f1120e, which is when the TMIO driver was introduced. I don't know enough about 3DS development yet to know what to do at this point, though.

As an aside, in previous commits, although checking FBI would not lead to a hang, all rxTools would output was:
```
Completed!
Region: USA
(A) continue
```
From reading the source code, it's not finding or it's failing to open the app (see  [rxTools/rxtools/source/features/downgradeapp.c:525](https://github.com/roxas75/rxTools/blob/d6cf5453ed8d594bc8d3e4a7c6100ffc37c48c33/rxtools/source/features/downgradeapp.c#L525), I think that's the conditional that's failing). If this is expected behavior for sysNAND, then so be it, but if not, I can open another issue to track this thing. (EDIT: Forgot to mention sysNAND has the original Health & Safety app, not FBI, so I would think it would be detected...).

173210:
This pull request fixes #358.


Syphurith:
It seems Unicode or ANSI made some severe problems.. It should be periodically moved to Unicode..

173210:
@Syphurith It's just a bug in downgradeapp, not related to Unicode.

Syphurith:
@173210 Sorry i didn't find that before reading the code. *I found no email from your github.io site..*

173210:
@Syphurith I found also Unicode related bugs LOL
The fixes don't seem enough to fix the bug. I'm still investigating...

Syphurith:
@173210 Don't push yourself too hard. There might be plenty of them.. *Orz*.

173210:
88f16f760a88f5bead1f75c56aa1085c77c4eb69 is the very bug causing #358.
The size of the struct was 4 bytes smaller, so `memset` and other functions had overwritten something else.

@Syphurith Thank you for advice. I forgot to do that :P
c24b0d1ad8b670fad3ad419d652b7dcf3f64f779 is a fix for all printf formats.

Syphurith:
@173210 Thanks this is much better now. Eh..
I've seen the change inside tmd structure.. so we were using wrong structure before (orz).

Desterly:
configuration.c is a mess too I believe.  probably the same issues here would apply there and fix the toolset problem

urherenow:
Toolset problem?

Desterly:
er dataset.   I think it's minor though.. it copies the firmware now but not creating the config file yet

PR #371, opened 2016-01-27
downgradeapp.c - update 'fbi' to unicode

Desterly:
This resolves injecting FBI for me at least.  Otherwise it would just hang


mariogamer2:
Some user report that fe18563adc81945b1f80210a57268f46a36c0566 block on "backink up Health and Security",does this pr fix this?

Desterly:
that's the last thing I would see on mine... it get down to the later point where it say it was injecting and would just hang... so yes, it very well may

PR #389, opened 2016-02-04
Fix Chinese translation

mxalbert1996:


Syphurith:
Could you please tell me what you did exactly in description of this pull request? Is some of them fixed and some useless entries removed? *Tab and space could sometime be a problem, just check the `File Changed` page in this pull..* And happy new year festival!

mxalbert1996:
@Syphurith Well I fixed some inaccurate and wrong translations. Just changed to Tab to match the English lang file. And Thanks!

urherenow:
Looking at this file and change confuses the hell out of me. I can't tell when and where %ls is needed. Some lines have it. Some lines don't. Some have it after the kanji. Some have it in the middle (kanji both before and after). Some have it multiple times in a row...

Can someone please explain or at least post a link to a good doc about it?

mxalbert1996:
@urherenow `%ls` represents the string argument used to fill in the sentence. For the details of `swprintf` function and the specifiers you can look at http://www.cplusplus.com/reference/cstdio/printf/ and http://www.cplusplus.com/reference/cwchar/swprintf/. Maybe looking at the `en.json` would be easier to understand.

RandomIntermition:
not 100% sure but i'm pretty sure LINE 65 says 'reservation/retention'? wouldn't a word for save/preserve make a better translation? or are retain and retention the same in simp'chinese? :S
just a minor thing, can still be understood i know but 'the better the better' right ;)

mxalbert1996:
@RandomIntermition Thanks for advice :)

RandomIntermition:
yes, that sounds better now.
had a second look top to bottom and didn't notice any other 'errors'. :)

PR #404, opened 2016-02-10
guide update

RandomIntermition:
updated guide in ref. to n3ds current method (beta)

also renamed old 3.0 to alpha, to avoid it being seen as current version.


AlbertoSONIC:
Error #404: Pull Request not found xD

Btw, what does merge remote-tracking branch do?

RandomIntermition:
that's because there where updates to main branch midway between this pull request and my fork, i should of updated the fork then made the edits but i did them both at the same time.
(so basically it's all of the changes that happened to rxtools master vs my master, changes that have actually already been applied)

to make things simpler i'll close and re sync everything then issue it again.

PR #408, opened 2016-02-11
docs update

RandomIntermition:
This is a reup of the previous pull request #404 but with the sync error removed.

info:

removed a few files that aren't needed,
the docs folder not consists of ONLY .pdf documents and an editors notes to allow future editors to keep maintain the same formatting.

changes:
quickstartQuide updates to match BETA and n3ds keys info.
makefile updated to use BETA pdf
older pdf renamed to alpha to avoid confusion.

files removed:
two svg image files that already exist within the docs and are no longer needed.
and an incredibly outdated edit of rxtools.pdf that is no longer needed.
current .pdf docs can be easily edited so additional formats being kept on git docs just causes clutter.

additional:
rxtools.pdf has not been updated to reflect 3.0 yet.


urherenow:
I don't understand the need to keep an older .pdf. If someone is installing an older rxTools, they will HAVE the older .pdf that comes with it anyway.

?

RandomIntermition:
no they won't, the older .pdf was only included in 1 release/handful of nightlies. so anyone working with previous versions that didn't include it in build may still want it.
especially seen as changes in the newer version only apply to n3ds users.
(if you read the notes you'll see it's being left there for ref. until a final 3.0 release)

only the current version is included in make so unless you're someone working on docs it really shouldn't be an issue.

it's the equivalent of saying "we should remove all prior releases because there's a new release out, people who have older releases will have the files so don't need to keep them".

Syphurith:
~~I hope the file `QuickStartGuide_v3.0.pdf` or `QuickStartGuide_v3.0(BETA).pdf` could be a more stable name `QuickStartGuide.pdf` that without any version marks. Git would also deal with the document file, so just update the `Document Release Date` section in document when it is updated.~~

RandomIntermition:
Cancelled temporarily.
Awaiting confirmation on an issue brought up by @urherenow relating to n3ds firm setup.

PR #424, opened 2016-02-16
docs update make unchanged

RandomIntermition:
no build related edits have been made so can be merged without errors.

single line edit in QSG, deleted an old ref. to firmware.bin in QA.
title remains the same so won't affect make.

added a first of the updated version of what was rxtools.pdf
it's now named rxtools_manual with the inner title of "Full Manual".
it isn't finished but thought i'd upload an edition just so there's something there rather than that old file (rxtools.pdf) which i've deleted in this pull.

to avoid confusion final doc name has been appended with DO NOT USE, so no one adds it to make.

EDIT: the rest of the comment has been removed as it was just talking about unfinished work in main man .pdf which is now 3/4 of the way to being finished.

**try to get this merged by the end of the week, then i'll have finished the first revision and will submit for review, so the community can open an issue and point out what i'm sure will be many many mistakes that have gone unnoticed.**


RandomIntermition:
old

PR #440, opened 2016-02-26
Update Makefile

b-:
Added export PATH	:= $(DEVKITARM)/bin:$(PATH) from the CakesROPSpider Makefile. This solves some compilation issues I was having on OS X 10.11 El Capitan.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-
+export PATH := $(DEVKITARM)/bin:$(PATH)
 CODE_FILE := code.bin
 SYS_PATH := rxTools/sys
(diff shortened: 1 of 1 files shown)


b-:
The path issues I was having are *probably* user error on my part, but CakesROPSpider's Makefile still prevents such user errors as long as $DEVKITARM is defined properly. By adding this line to the main rxTools Makefile we're actually ensuring that $DEVKITARM/bin is the first folder in the command search path, which means that anybody with a properly installed copy of devkitARM needn't add it to their path variable for compilation.

Also, the fact that compiling requires `make release` was getting annoying to me, since most Makefiles I've encountered have a sane default target. By listing `all-target-brahma` in the Makefile as the first target whose name doesn't begin with a `.` (e.g., `.PHONY` — what is that anyway?) it was the *only* target that would be compiled upon simply typing `make`, requiring a user to type `make release` to properly compile rxTools. `.DEFAULT_GOAL` is a Makefile variable pointing to what the default target *should* be (in this case, `release`).

PR #450, opened 2016-02-29
rxtools: Add init

173210:
init is an alternative for bigpayload in CakeHax and can be an entry point for delebile's arm9loaderhax.


urherenow:
if this is what I just tried, it does not work. The current roxas75/rxTools arm9loaderhax branch at least loads the menu (but data installation suite is broken).

173210:
It needs modifications for arm9loaderhax. Please consider it's just for CakeHax for now.

AlbertoSONIC:
Is this really needed? rxTools arm9loaderhax branch already works fine in both arm9loaderhax and other entrypoints, using cakehax bigpayload and modded makefile which outputs arm9loaderhax.bin (which basically is code.bin with first 0x12000 part removed)

173210:
Yes. arm9loaderhax branch will not work with CakeHax and possibly causes unexpected results.

173210:
Ok, I read again and found it has a dirty hack to extract bigpayload from CakeHax. So it may work. But still it's a hack and can be corrupted.

AlbertoSONIC:
Well if you think doing a separate init is better then my dirty hack then I'll merge this... But please fix arm9loaderhax support ;)

AlbertoSONIC:
Also, rxTools will be renamed to RX Custom Firmware so maybe tools.bin won't fit that much

173210:
Originally init was developed because `bigpayload` sets MPU up incorrectly for new 3DS. Now we need to update MPU configuration in the initial bootstrap.

173210:
> RX Custom Firmware

Wow.

AlbertoSONIC:
Mmm so now code.bin is the init and tool.bin is the actual RX code?

173210:
Yes.

AlbertoSONIC:
So code.bin is what will be renamed to arm9loaderhax.bin and placed in sd root...

173210:
No. `code.bin` includes ROP, but `arm9loaderhax.bin` shouldn't.

AlbertoSONIC:
init.bin?

urherenow:
You told me earlier that it was the same thing... Is that why it didn't work?

173210:
@urherenow First, it will generate `bin/init.bin` (previously `rxtools.bin`), then `bin/tool.bin`. `bin/init.bin` will be assembled with CakeHax. The result is `code.bin`.

RinMaru:
wait your not planing on making rxcfw a9lh only right?

173210:
@Zetchzie This pull request and `arm9loaderhax` branch won't kill other entry points.

RinMaru:
thank god im not ready to set a9lh as my default.

urherenow:
Before everything is working AND emunand9 and decrypt9 can be launched directly from the RXcfw, AuReiNand is better for a9lh anyway. It already works, and the sysnand launch with 9.0 native_firm also keeps it compatible with Gateway.

mariogamer2:
@AlbertoSONIC Maybe a name more than this XD

173210:
77512fe3db1a1950622849f23e64f2db3ae82d80: Updated the MPU configuration according to delebile's arm9loaderhax.
https://github.com/delebile/arm9loaderhax/blob/d52f7b46ec734d7a3390e098883872336ed0481b/payload_stage1/source/_start.s

Probably it boots. But no new 3DS support. But I have an idea to fix that.

RinMaru:
so nightlys are no longer a thing its been canceled. when is the next beta expected?

AlbertoSONIC:
@173210 Cakes has an MPU settings code which works on both arm9loaderhax and other entrypoints. It's also working on new 3ds... You may take a look at it (it's the one I used in arm9loaderhax branch, it works)

